PR #19, opened 2016-04-12
Table page break bug and PDFBOX character bug

rototor:
This is a reduced testcase which shows two bugs:
- The repeated table header is broken on page break. openhtmtopdf shares this bug with flyingsaucer.
- The arrow symbol is not correctly resolved from the unicode
  character. Instead a # character appears. There are even two # characters on places were no character is shown... If you view the html in a browser you see the right symbols. This works fine in flyingsaucer - after I manually register an other font (e.g. Noto Sans) which has this unicode symbol, as the fonts included in iText 2.1.7 don`t have that character.

Do you know how this "table repeated headers on page break" stuff works? I tried to figure it out / debug it, but I just don`t really understand where the repeated table headers are created. Somehow the top-margin is not calculated correctly after the page break. If you increase the margin-top in line 126 of the html file you can see that the header somehow gets more into the view. If you can give me some pointers were to look I can try to future investigate that problem. But at the moment I really don`t know were to look.

I don`t think that examples is the right location for such a test case. But its not a unit case neither. Please tell me were I should place this files.

Thanks for reviving flyingsaucer under a new name! Beside some rare bugs like these its a very useful library to create reports (using html). And also to render html snippets to images (which I also use in my projects)

### DIFF
--- a/openhtmltopdf-examples/src/main/java/com/openhtmltopdf/testcases/RepeatedTableSample.java
+++ b/openhtmltopdf-examples/src/main/java/com/openhtmltopdf/testcases/RepeatedTableSample.java
@@ -1,0 +1,33 @@
+package com.openhtmltopdf.testcases;
+
+import com.openhtmltopdf.bidi.support.ICUBidiReorderer;
+import com.openhtmltopdf.bidi.support.ICUBidiSplitter;
+import com.openhtmltopdf.pdfboxout.PdfBoxRenderer;
+import com.openhtmltopdf.pdfboxout.PdfRendererBuilder;
+import com.openhtmltopdf.pdfboxout.PdfRendererBuilder.TextDirection;
+import org.apache.pdfbox.io.IOUtils;
+import org.apache.pdfbox.util.Charsets;
+
+import java.io.FileOutputStream;
+
+public class RepeatedTableSample {
+	public static void main(String[] args) throws Exception {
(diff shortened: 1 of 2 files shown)


danfickle:
Thanks @rototor - I'm not sure why font fallback is not working - I did test it. However, I'll take another look tomorrow. As to the table header issue, as you've no doubt found, the layout code is highly complex, but I'll take a dive into that too. I think examples is a reasonable place for test cases as it has all the required imports.

danfickle:
@rototor - I changes a couple of lines (and added the `NotoSans-Regular.ttf` file) to your test case and got the arrow printing out in a PDF.

The lines were:

``` css
src: url("NotoSans-Regular.ttf") format("truetype");
/* We don't support format at this time so remove it: */
src: url("NotoSans-Regular.ttf");
```

``` java
builder.withHtmlContent(html, "");
/* We need a base url to resolve fonts so: */
builder.withHtmlContent(html, RepeatedTableSample.class.getResource("/testcases/").toString());
```

Let me know if you still can't get the arrow printing. I'm investigating the table header issue now.

rototor:
@danfickle After I copied the Noto-Sans Font into the testcase directory, I see the arrows. But there are still # characters in the pdf - on every page in the left top corner of the "body" area. Where every they come from.... no idea.

danfickle:
I think the table break bug is to do with the `ContentLimitContainer` getting the wrong page content limits, I'm just not sure how this is possible.

Issue #20, opened 2016-04-15
Use custom font


aleksandr-m:
Any way to use custom font? There is `PdfBoxFontResolver`, but currently don't see how to add custom font to it.

danfickle:
Currently, the easiest way is to use a font-face rule in your css. Note: Only truetype fonts are supported and this will not currently work in the Java2D renderer - see #17

``` css
@font-face {
    font-family: "Noto Sans";
    src: url("NotoSans-Regular.ttf"); /* Use an absolute or relative url. */
}

/* Later, use font as named: */
body {
    font-family: 'Noto Sans', sans-serif;
}
```

If you want to add it in code instead, let me know and I'll give you a code sample.

aleksandr-m:
The thing is I need to load font from the jar. I can get `PdfBoxFontResolver` from `PdfBoxRenderer` but currently `addFont` method loads font with `FileInputStream`. Can you add another `addFont` method which loads resources from the classpath (e.g. with `getResourceAsStream`)?

danfickle:
I think the referenced commit does what you want. You can also use a jar url or a relative url if the document url is from a resource. Let me know if the provided method doesn't work. Thanks, Daniel.

aleksandr-m:
Initially I was thinking too about adding method with input stream but then noticed that there is check for `.ttf` and opted for `TrueTypeFont`.

Anyway, now it should work. Thanks.
I'm planing to upgrade [struts2-pdfstream](https://github.com/aleksandr-m/struts2-pdfstream) when you release next version with this improvement.

PR #22, opened 2016-04-19
Allows to add existing font to font resolver

aleksandr-m:
This should solve issue with custom fonts. See #20


danfickle:
I didn't use this pull-request directly, as I didn't think exposing TrueTypeFont from font-box was a good idea. However, I have added a method to add a font based on an input stream which you can get from a resource. Thanks again, Daniel.

PR #31, opened 2016-07-13
Implement -fs-page-break-min-height

rototor:
This CSS property allows to force a page break before an element if not the specified space is left on the
current page.

CSS orphans and page-break-inside:avoid are nice, but dont work always,
especially with complex tables.

This CSS property allows to workaround many page break problems.

Also adds a sample document which demonstrates the effect of
this property. Also the TestcaseRunner now prints onto the System.out
were the output PDF was written.

@danfickle I'm not sure if -fs-page-break-min-height is the best name for this property. If you can come up with some better name, I'll change it.

Yes, this is some kind of "workaround" hack. But I "need" openhtmltopdf work for me now, as I want to redo some complex JasperReports reports with FreeMarker + openhtmltopdf. Of course it would be better to just fix the layout bugs, but as seen in #19 its not that easy :(

If you want to see what bugs I mean => just comment out the CSS properties in FSPageBreakMinHeightSample.html:176 and FSPageBreakMinHeightSample.html:173  and run the TestcaseRunner.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/css/constants/CSSName.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/css/constants/CSSName.java
@@ -272,7 +272,7 @@ public final class CSSName implements Comparable {
                     NOT_INHERITED,
                     new PrimitivePropertyBuilders.FSCheckboxStyle()
             );
-    
+
     /**
      * Unique CSSName instance for CSS2 property.
      */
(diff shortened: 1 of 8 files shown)


danfickle:
Thanks @rototor - The code looks good - The only queries I have are:
- I don't think this property should be inherited. Layout properties generally are not.
- Is it possible to make a simpler test case that just demonstrates this property?
- What happens if the figure provided is greater than the page height?

If you like I can take care of these minor concerns after the merge. Let me know if that is alright. Thanks again and sorry for the delay, I've had a cold.

rototor:
@danfickle to be honest, I just copied the CSS declaration and did not really care about the inherited specification. I think you are right that this property should really not be inherited.

The last point should not matter, as -fs-page-break-min-height is more or less nothing else than a "page-break-before:conditional". I.e. it only affects places were a page-break-before could cause a page break. So it would just add an other page break. But I have not tested this yet.

If you want to do this little changes, feel free to do so. I wont have time the next 4 days.

danfickle:
OK, thanks a lot @rototor

PR #32, opened 2016-07-13
Add visibility:-fs-table-paginate-repeated-visible for paginated tables.

rototor:
This allows to render some elements on the continuation of a repeated
table header on a new page. E.g. a span containing the word "continued".

For this to work the table must have -fs-table-paginate: paginate; set.

Using visibility the element is always correctly layouted, but only
painted after the table is broken into a new page.

@danfickle As with #31 feel free to suggest a better property name. 

A customer of mine wants to have "(Continued)" in the table header if the table is broken across pages. This implements it.

I will update the testcase in #31 to use this feature if #31 is merged.

While implementing this I found out, that the table header elements are special handled in the BoxCollector. There is some "magic" which makes them repeat. So  #19 might be just some wrong calculation in TableBox.updateHeaderFooterPosition(). I'll likely look into this in coming the next weeks.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/css/constants/IdentValue.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/css/constants/IdentValue.java
@@ -103,6 +103,7 @@ public class IdentValue implements FSDerivedValue {
     public final static IdentValue FONT_WEIGHT_900 = addValue("900");
     public final static IdentValue FS_CONTENT_PLACEHOLDER = addValue("-fs-content-placeholder");
     public final static IdentValue FS_INITIAL_VALUE = addValue("-fs-initial-value");
+    public final static IdentValue FS_TABLE_PAGINATE_REPEATED_VISIBLE = addValue("-fs-table-paginate-repeated-visible");
     public final static IdentValue GEORGIAN = addValue("georgian");
     public final static IdentValue GROOVE = addValue("groove");
     public final static IdentValue HEBREW = addValue("hebrew");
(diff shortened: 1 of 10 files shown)


danfickle:
Thanks @rototor - this looks good to go and should be quite useful. I'll merge it after #31 as you propose.

michaelborn:
I'm curious why the decision to implement this on the `visibility` property? Using `visibility` means that the element displays and renders (though invisible) even on the first page with the table header, thus displaying an empty space equal to the width of the element - see screenshot:

Screenshot_2019-09-03 bd2d27af-15a7-408d-8a17-901c33f4bdb3

and here's how the table header looks on subsequent pages:
Screenshot_2019-09-03 bd2d27af-15a7-408d-8a17-901c33f4bdb3(1)


To me, a more standard use case would be `display: -fs-table-paginate-repeated-visible;`, in which the `[Continued] ` text can hide completely on the initial page. This would keep the table header from showing an awkward empty space on the initial page.

rototor:
This is an implementation limitation. The repeated-header row elements are only repositioned "as-is" and repainted on new pages. No kind of layout takes place on page breaks. So the only thing you can do is to paint or not paint something. Which this visibility style controls. 

Sorry, you need to move the [Continued] after the text. Everything else wont work.

It might be possible to perform a layout of the table header row on every page break, but this will likely be very complex to implement. But feel free to try.

michaelborn:
Fair enough. I realized that placing it after would work well enough, and added some documentation for this in the Wiki. https://github.com/danfickle/openhtmltopdf/wiki/Page-features

PR #34, opened 2016-08-11
Fix hairline bugs in table borders.

rototor:
We need to get the full interior Area, otherwise we will have some
super small hairline gaps in the PDF. They are always small but also
always visible in the PDF independent of the zoom level.

Broken
image

vs.

fixed
image


danfickle:
Thanks @rototor - looks good. I'm sorry for going quiet on this project. I will get back to it soon.

PR #45, opened 2016-10-30
Started implementing transform/transform-origin. The CSS declaration

rototor:
is halve way parsed... Any visible effect on output is still missing.

This is a WIP, not nearly finished.


danfickle:
Thanks heaps @rototor

PR #48, opened 2016-11-01
Add support for max-width and max-height properties on img elements

achuinard:
This commit should allow PDF rendering to support max-width and max-height properties on img elements.

### DIFF
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxReplacedElementFactory.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxReplacedElementFactory.java
@@ -20,7 +20,11 @@
 package com.openhtmltopdf.pdfboxout;
 
 import com.openhtmltopdf.css.constants.CSSName;
-import com.openhtmltopdf.extend.*;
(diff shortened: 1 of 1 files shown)


danfickle:
Perfect! Thanks @achuinard 

Merging now.

achuinard:
Thanks. I thought that the best place to put this might have been in
BlockBox. Are images rendered as a BlockBox? Why didn't the max width and
max height properties get taken into account already?

danfickle:
Images are rendered as special replaced elements (like form controls). I think this was how it was specified in an old HTML standard. We might have to revisit this if we want to apply other CSS on the img element such as transform.

achuinard:
Sounds good. How does the release cycle work? Will RC8 be pushed to the repository?

danfickle:
I'd prefer to not make a release so soon after the last.
Can you get away with wrapping the image in a block?

``` html
<div style="max-width: 80px;">
  <img src="image.png" alt="title" style="width: 100%;"/>
</div>
```

achuinard:
Can you release the new version to Maven?

danfickle:
Hi @achuinard 
I've got some uncommitted code for CSS transforms that I'm working on. Can you wait a couple of days or do you need a release for Monday?

achuinard:
I can wait til end of week at latest.

Issue #23, opened 2016-04-22
Add support for SVG images


ngns:
Currently, SVG images are not supported. PDFBox does not "draw" SVGs natively. Best practice is, to use apache batik to transcode SVGs to PDFs and embed these as PDFFormXObject via PDFBox. 

_See also _
- SVGs via batic in PDFBox: https://www.mail-archive.com/users@pdfbox.apache.org/msg06998.html
- Transcoding SVGs with batic: http://stackoverflow.com/questions/6875807/convert-svg-to-pdf

danfickle:
I decided to create our own transcoder rather than the transcoder provided by FOP. This gives the benefits that it can output to Java2D as well as pdf and can use our user agent, font resolver, etc. However, it is a bit more work so I'm still debugging it and will commit something tomorrow for testing.

ngns:
Should we keep the transcoder to use flexible? Batik is very mature, and will likely be quicker to implement for complex SVGs.
If your are going for Java2D target, maybe it could be built upon existing solutions like Flamingo SVG Transcoder (also batik-based)? http://ebourg.github.io/flamingo-svg-transcoder/

danfickle:
This is a screen shot showing simple SVG graphics in a PDF.
![openhtmltopdf-screenshot](https://cloud.githubusercontent.com/assets/1415728/15985052/54e06146-3024-11e6-9c59-50dd6d84cd51.png)

danfickle:
Todo for proper SVG output:
- [x] Basic fill and draw operations
- [x] Basic text output
- [x] Inline SVG in HTML document
- [ ] External SVG in image
- [ ] Linear gradient support
- [ ] Alpha support
- [x] Custom font support
- [ ] Insert raster image in SVG
- [x] SVG transform support
- [ ] SVG clipping support
- [ ] Testing

danfickle:
With commit bb586a9 you can output custom fonts in SVG like this (text color is still wrong and gradients aren't implemented):

![svg-custom-font](https://cloud.githubusercontent.com/assets/1415728/16080350/b60c19a4-334b-11e6-9bda-d7718f681045.png)

``` html
<html>
<head>
<meta charset="UTF-8"/>
<style>
@font-face {
  font-family: 'handwriting';
  src: url(JustAnotherHand.ttf);
}
svg {
    display: block;
}
body {
    font-size: 50px;
}
</style>
</head>
<body>
[...]

PR #66, opened 2017-02-02
#23: PDFBox-Graphics2D based SVG rendering

rototor:
This implements the SVG rendering using my pdfbox-graphics2d (see https://github.com/rototor/pdfbox-graphics2d). 

Run the testrunner and look at the resulting svg-inline.pdf. The SVGs are drawn, but the position and size is wrong. Also the text on the barchart-sample is not displayed. (This works in the pdfbox-graphics2d testdriver, so something must be wrong with the font resolving, did not investigate yet). 

At the moment I have no idea how to position right.

The PDFBox-Graphics2D puts all drawcalls into a XForm. This XForm can then be placed wherever one may want. 

PdfBoxOutputDevice.drawWithGraphics() and PdfContentStreamAdapter.placeXForm() may need to be adapted to scale and place the resulting XForm correct. 

drawWithGraphics() should not only work with SVGs but in general, because we have some reports were we draw bar graphs using Graphics2D. These reports are still using JasperReports (and iText underneath), but we want to migrate away from  JasperReports and iText to openhtmltopdf.

It would be perfect if one could call drawWithGraphics() only with the position arguments of the ReplacedElement and the resulting XForm would be placed correctly. 

Do you have any hints how this can be implemented or could you fix the positioning?

Thanks.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/OutputDevice.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/OutputDevice.java
@@ -111,4 +111,9 @@ public interface OutputDevice {
     
     public boolean isSupportsCMYKColors();
 
+    /**
+     * Draw something using a Graphics2D at the given rectangle.
+     */
+    public void drawWithGraphics(float x, float y, float width, float height, OutputDeviceGraphicsDrawer renderer);
+
 }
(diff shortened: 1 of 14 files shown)


danfickle:
Thanks @rototor 
Just letting you know I'm still here and will take a look in the next couple of days.

danfickle:
Thanks again @rototor - the code here and in your repository looks excellent. I'll merge now and have a play around.

Regards,
Daniel.

PR #72, opened 2017-02-19
Pdfbox graphics2d fixes

rototor:
Just some small fixes, the green debug rectangle is not needed any more. And the old (and now obsolete) Graphics2D adapter is also not needed any more. And I released a new version of pdfbox-graphics2d. It can now also handle TexturePaint.


danfickle:
Awesome! Thanks again @rototor

Issue #63, opened 2017-01-15
Possible bug in Flying Saucers 'Graphics2DRenderer'


skjardenCode:
Hi Daniel,

thanks a lot for your ongoing efforts on this helpful project!

We recently came up with the use case to render a certain small HTML to an image before we render it to PDF. As OpenHtmlToPdf is based on Flying Saucer, we tried to use the ```com.openhtmltopdf.simple.Graphics2DRenderer``` class to do so, which should be an easy task.

Unfortunately (using Java 6 and Windows 10 with the latest OpenHtmlToPdf version available in Maven Central), the image output of ```Graphics2DRenderer``` yield wrong results, mainly what seems like an encoding problem.

Please, see the following simple example code:

```java
String xhtml = 
        "<!DOCTYPE html PUBLIC \"-//W3C//DTD XHTML 1.0 Strict//EN\" "
                + "\"http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd\">"
        + "<html xmlns=\"http://www.w3.org/1999/xhtml\" xml:lang=\"de\" lang=\"de\">"
            + "<body>"
                + "Moonbase"
            + "</body>"
        + "</html>";

InputSource inputSource = new InputSource( new BufferedReader( new StringReader( xhtml ) ) );
Document document = XMLResource.load( inputSource ).getDocument();

Graphics2DRenderer graphics2dRenderer = new Graphics2DRenderer();
graphics2dRenderer.setDocument( document, null );

BufferedImage bufferedImage = new BufferedImage( 500, 500, BufferedImage.TYPE_INT_RGB );
Graphics2D g2d = bufferedImage.createGraphics();

graphics2dRenderer.layout( g2d, new Dimension( 500, 500 ) );
graphics2dRenderer.render( g2d );

g2d.dispose();

File outputfile = File.createTempFile( Graphics2DRendererTest.class.getSimpleName(), ".png" );
ImageIO.write( bufferedImage, "png", outputfile );
Desktop.getDesktop().open( outputfile );
```

The image you get (trying to render the word "Moonbase") is the following:

![image](https://cloud.githubusercontent.com/assets/23582335/21961844/2294ffc0-db15-11e6-8f2e-df7f942aafad.png)

We than integrated the latest Flying Saucer version into our project, next to OpenHtmlToPdf, and just changed the line

```java
import com.openhtmltopdf.simple.Graphics2DRenderer;
```

to

```java
[...]

Issue #73, opened 2017-02-23
Is there a way to output an image file rather than a pdf?


Jo3McCarthy:
Hi, All;  

I think the process of rendering has some interim processing steps where content is converted into images.  Is there a method to output the images directly?  Or, could you please point me to a  point in the code where I could implement that function?

Thanks,

Jo3

danfickle:
Hi @Jo3McCarthy 

There is a Java2D output device in the core. See the swing package I think. Unfortunately, it is broken at the moment. Also see #63 
If you'd like to work on it, I'd be happy to help. I think as a start we should work on moving it to its own module.

Regards,
Daniel.

rototor:
Just saw this issue. With #74 I started to implement something similar to the PdfRendererBuilder. If you just want to quick render some HTML as Image you could use Graphics2DRenderer.renderToImageAutoSize(). But you need a URL (e.g. file.toURL().toExternalURI()) for that to work. With the Java2DRendererBuilder (name can still change) you can also render in memory HTML string as image. In the midterm I think we should get rid of the old Graphics2DRenderer and Java2DRenderer. The later does not even work at the moment.

danfickle:
OK, I've implemented a new renderer and worked on the builder by @rototor (in a copy) to output to paged images. One can incorporate the java2d sub-module and use it thus (but it is likely to change):
````java
		BufferedImage bf = new BufferedImage(1, 1, BufferedImage.TYPE_3BYTE_BGR);
		final Graphics2D g2d = bf.createGraphics();
		
		Java2DRendererBuilder builder = new Java2DRendererBuilder();
		
		builder.withUri("file:///Users/me/Documents/pdftests/image-output-device-test.htm");
		builder.useLayoutGraphics(g2d);
		builder.usePageProcessor(new DefaultPageProcessor("/Users/me/Documents/pdftests/images/%page-number%.png"));
		builder.useSVGDrawer(new BatikSVGDrawer());
		
		Java2DRenderer renderer = builder.buildJava2DRenderer();
		
		renderer.layout();
		renderer.writePages(0);

		g2d.dispose();


````

which will output a PNG image for each page of the document in the correct size.

danfickle:
Start of a checklist for the Java2D (or any new) output device:
+ [x] Output simple raster images
+ [x] Output simple SVG images (with no text or embedded images)
+ [ ] Output raster images referenced by data URIs.
+ [x] Import font-face rule fonts.
[...]

PR #74, opened 2017-02-26
Graphics2d builder and testdriver 

rototor:
I've started to update the Graphics2D support. To do so I've created a Java2DRendererBuilder and called it in the testdriver to write a PNG output beside the PDF output. I also tried to reproduce #63  with this, but I could not. 

The PNG currently shows some missing features and bugs:
 - CMYK colors are not supported at the moment. We should just use some simple hardcoded CMYK->RGB conversion for that as a start. I think using some kind of ICC-based color conversion is overkill here.
- For whatever reason the text in the svg-sample is rendered broken. I think for Java2D the correct fix is to just use TextLayout and let TextLayout do all the rendering. Then Java2D will have as much Bidi support as Java itself has. 
- Also the (partly) unused support for drawing selections is obsolete. If someone wants to display HTML in his application he will likely rather use the JavaFX-WebKit component from the JDK. For me the Java2D support is mostly the ability to render HTML to an image, so selections aren't really needed. I think the selection support should just be deleted.
- The current place where I've put the Java2DRendererBuilder is not the best one. Also the innerclasses should maybe be outerclasses in some other packages. Feel free to suggest a better place.

When I find time again I would like to work on this things. Do you have any objections?

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/OutputDevice.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/OutputDevice.java
@@ -30,17 +30,8 @@ import java.awt.geom.AffineTransform;
 import java.util.List;
 
 public interface OutputDevice {
-
-	// Required for SVG output.
-	public void saveState();
-	public void restoreState();
-	
 	public void setPaint(Paint paint);
 	public void setAlpha(int alpha);
-	
-	public void setRawClip(Shape s);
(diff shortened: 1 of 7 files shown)


danfickle:
Hi @rototor 

Thank you again, this looks really good. I'm going to merge now and then I'll try to put the Java2D output related code in its own sub-module. We can then continue the discussion at #73 

Regards,
Daniel.

PR #77, opened 2017-03-12
Drop simple.Java2DRendererBuilder and implement single paged output

rototor:
Drop the Java2DRendererBuilder in the core package and use the java2d
package. Also provide a BufferedImage PageProcessor.

Allow to output everything as single page. And adapt the testdriver to
use both modes.

Note: I am not sure if the output to file should be the DefaultPageProcessor. Everytime I need to output to an image using java2d I want to draw to

 - Another graphics2d
 - or a BufferedImage, which is then encoded and sent over network (as respone to a HTTP request)

So a default which write files somewhere seems "strange" for me. But as I can just change it, its no problem. I am not sure if the way I disabled the paging is the right one, but the output mostly seems sane (only the FSPageBreakMinHeightSample-sample seems strange in single page mode).


danfickle:
Thanks again @rototor 
You're really pushing this project along! As for the output stream supplier for default page processor, I thought most people would get the bytes using a byte array output stream, but you're right, I didn't think about those who need to post process a buffered image. We should probably rename the default page processor to reflect its output stream nature.

PR #81, opened 2017-04-07
Support all <meta>-data

rototor:
Put the content of all non standard metadata into the Custom-Metadata of
the PDF. This allows to e.g. insert debuginfos into the PDF and also to
create additional metadata for other tools.


danfickle:
Thanks @rototor 
Down the track we should reimplement support for xmp metadata, but this is a good solution for now.

PR #83, opened 2017-04-09
UNICODE font justification support

backslash47:
Text justification implemented in https://github.com/danfickle/openhtmltopdf/issues/3 will not work if using UNICODE fonts. 

The reason is that Tw parameter (word spacing) is ignored for non simple fonts as defined in pdf standard at https://www.adobe.com/content/dam/Adobe/en/devnet/acrobat/pdfs/PDF32000_2008.pdf#651 page 253.

Therefore I reused the code from old iText output device which solved this problem.


backslash47:
The resulting pdf will work only in some pdf viewers. I have prepared a solution, but unfortunatelly it needs support from pdfbox. Therefore I made a pull request for pdfbox.

danfickle:
Thanks @backslash47 for looking at this. Hopefully, you can provide another pr when supported by pdfbox. Thanks again.

PR #84, opened 2017-04-11
Updated scaling logic for max-width and max-height images

achuinard:
I ran into some issues when my images height and width both exceeded my max-height and max-width properties. I believe this logic fixes it. The only commits which are new though are the ones from April 10, and this says that my Nov 1 and Nov 2 commits are also being pulled in...but those were pulled in already. Please confirm I didn't do this wrong.

Thank you.


danfickle:
Thanks @achuinard 
I'm not sure of the finer details of pull requests but the changed files section looks correct so I'll squash and merge. Down the track we should think how to apply this code for all replaced elements.

PR #90, opened 2017-04-21
Add possibilty to programatically disable logging before first log

GrammyTraore:
Hello!

I was using your nice lib and found out that calling `XRLog.setLoggingEnabled(false)` before any log was not working. In fact the value was always overriden by the configuration or default value when XRLog gets intialized at first log (maybe it's wanted though...)

Here is a quick fix I made, hope it can help!


danfickle:
Thanks a lot @GrammyTraore 
That should do the trick.

PR #95, opened 2017-05-09
Links in PDFs: Track link overlapping per page, not for all pages

rototor:
Currently the area occupied by a link is tracked in a Set to avoid
putting two links above each other. No idea why this can happen normaly,
but it least it tries to avoid that. But this Set is not per page,
its global for all pages.

Guess whats happend in a report on the production system of a customer
of mine :)

They have tables spanning many pages, and magically all links which happen
to be exactly at the same position like a link on another page the link
does not get exported.

With this patch the Set is tracked per page.

Is this Set really needed? Why was it introduced in the first place?

### DIFF
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxLinkManager.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxLinkManager.java
@@ -1,14 +1,11 @@
 package com.openhtmltopdf.pdfboxout;
 
-import java.awt.Rectangle;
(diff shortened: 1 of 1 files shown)


danfickle:
I think it was introduced because a link can contain multiple boxes, for example, if part of the link is bold. Because ````checkLinkArea```` is called for every box, you would generate multiple links for a single anchor element. Perhaps, the check should be that one ````a```` element only generates one link in the PDF rather than an area check. You'll notice this is how I did it for form handling.

The other problem with our link handling is that it can only generate a single rectangle. This won't work if the link goes over multiple lines for example.

Anyway, thanks a lot for for the correction and I'll merge now.

Issue #99, opened 2017-05-31
SVG rendering issues with batik-1.9


kneovector:
I found that the same SVG code that renders fine with batik 1.8 does not render properly with batik 1.9. No errors are visible in logs but the rendering is incomplete. I need to integrate with an existing project that uses batik-1.9 jars. Is there a known incompatibility with batik 1.9?

Output with batik-1.8 (correct):
<img width="610" alt="screen shot 2017-05-31 at 5 01 19 pm" src="https://cloud.githubusercontent.com/assets/1478145/26630073/defcb784-4622-11e7-94a2-69b35ff0514f.png">

Output with batik-1.9 (incomplete):
<img width="633" alt="screen shot 2017-05-31 at 5 01 27 pm" src="https://cloud.githubusercontent.com/assets/1478145/26630072/defb52ae-4622-11e7-8f59-47cd0e046a1d.png">

danfickle:
Hi @ketanmpandya 

The project hasn't been debugged with batik 1.9, although it looks like it could be a clipping problem. I'd suggest using the maven shade plugin to repackage batik in a custom svg-support pom. The closest I could come to find a guide on how to do this is at the below link.
https://www.elastic.co/blog/to-shade-or-not-to-shade

If you do get this working (or not), please let us know so others can learn. Good luck.

rototor:
@ketanmpandya This is rather an issue with pdfbox-graphics2d than with openhtmltopdf, as the SVG rendering runs through pdfbox-graphics2d. I've opend an [issue](https://github.com/rototor/pdfbox-graphics2d/issues/2) and will investigate whats going wrong here.  A sample SVG file to reproduce the bug would awesome.

rototor:
@ketanmpandya I can confirm that upgrading batik to 1.9 breaks the SVG rendering with pdfbox-graphics2d.

rototor:
I've created the pull request #104 which fixes that bug. My PDFBox-Graphics2D adapter had a bug with getClip() and clip(Shape). I've released a fixed version 0.5. You can wait till @danfickle merges this pull request and releases a fixed version or you can just add this dependency to your project

```
<dependency>
	<groupId>de.rototor.pdfbox</groupId>
	<artifactId>graphics2d</artifactId>
	<version>0.5</version>
</dependency>
```

PR #104, opened 2017-06-26
#99 Upgrade PDFBox-Graphics2D to bugfixed version 0.5. And upgrade

rototor:
Apache Batik to current version 1.9.


danfickle:
Thanks mate. I might try to do some more work on flowing columns before performing a release if that's ok?

rototor:
Of course, I'm not in a hurry

PR #108, opened 2017-07-20
ability to run examples from maven

jartysiewicz:
mvn exec:java -pl openhtmltopdf-examples
output is set to 'target/examples'


danfickle:
Thanks a lot @jartysiewicz - merging now.

PR #121, opened 2017-09-04
Use fonts for graphics 2d

rototor:
This pull request does the following things:

- Upgrade PDFBox to 2.0.7
- Upgrade ICU to 59.1
- Upgrade PDFBox-Graphics2D to 0.7
- Try to use fonts when rendering text with drawWithGraphics().

Should I split the version changes into a separate pull request or is this ok?


danfickle:
Thanks a lot @rototor

Issue #92, opened 2017-04-28
Add a way to render directly to PDF/A 1-3


schmitch:
Currently it would be great if there was a way to have a PDF/A-3 output without calling into PDFBox.

ebruchez:
@schmitch I am curious to know if you found a way of creating PDF/A with Open HTML to PDF?

schmitch:
@ebruchez sadly without patching openhtmltopdf it would not be possible.

Currently the only way to do it is to change the font loading to use `PDTrueTypeFont` instead of `PDType0Font` and use the WinAnsi Encoding. After that some needs to add a method, so instead of `run()` or `createPDF()` the `PDDocument` needs to be returned and metadata needs to be added.

ebruchez:
@schmitch Thanks, those are good pointers though.

harbulot:
Sorry, I don't know much about PDF/A, but I've noticed that the patch in PR #131 seems to break using CJK fonts.

I don't know whether [this validator](https://www.pdf-online.com/osa/validate.aspx) is correct, but I've tried a PDF generated with the latest commit (0c76faba77ee1735cd823281121273731e6d6049), without this PR, and it passes the validation.

This is the test, and I load both `OpenSans-Regular.ttf` and `cwTeXFangSong-zhonly.ttf`:

```html
<!DOCTYPE html>
<html>
<head>
<style type="text/css">
* {
    font-family: 'Open Sans', cwTeXFangSong;
}
</style>
</head>
<body>
    <h2>Test: cwTeXFangSong</h2>
    <div style="width: 80%; border: 1px solid black;">
        <p>含有中文字符的html输出pdf有乱码</p>
        <p>有支持中文的计划吗？</p>
    </div>
</body>
</html>
```

This is what Adobe Reader sees for the embedded fonts:

![image](https://user-images.githubusercontent.com/80994/30607812-35b26e64-9d6e-11e7-8fd8-1cd04a775177.png)

Just out of curiosity, what's wrong with these fonts?

schmitch:
the online validator is broken you either need to use https://github.com/veraPDF/veraPDF-library or have a Adobe Pro License. currently with PdfType0Font the CIDSet is incorrect for PDF/A actually you also need to change openhtmltopdf to render to a PDF/A

danfickle:
[...]

Issue #129, opened 2017-09-15
含有中文字符的html输出pdf有乱码


ReliefZk:
有支持中文的计划吗？

harbulot:
(Although I understand it's not always easy, it would help if you wrote these issues in English.)

Google Translate says that this is what you've written means:

> Html output with Chinese characters pdf has garbled
> 
> Is there a plan to support Chinese?
> 

It works, but you need to make sure the font you're using supports the characters you're using.

I'm not sure where to find TTF fonts with Chinese characters that you'd like to use, but Google seem to have a few available here: https://fonts.google.com/earlyaccess
(This is just an example, check if the licences are acceptable for your usage purposes.)


Here is an example (with `cwTeXFangSong-zhonly.ttf`, but it could be something else):

```java
PdfRendererBuilder builder = new PdfRendererBuilder();
// ...
builder.useFont(new FSSupplier<InputStream>() {
    @Override
    public InputStream supply() {
        return new FileInputStream("cwTeXFangSong-zhonly.ttf");
    }
}, "cwTeXFangSong", 400, FontStyle.NORMAL, true);
```

Then, in your HTML/CSS:

```css
html {
	font-family: "cwTeXFangSong", "Open Sans";
}
```

(You can use multiple fonts in the `font-familiy` declaration, and it will fall back to the next font in the list if it can't find the glyph to be drawn in the font.)

danfickle:
Thanks @harbulot 
Also see #115 for a complete example of using CJK fonts.

PR #131, opened 2017-09-19
use true type font loading instead of PDType0Font

schmitch:
actually this is important to actually have PDF/A support.

See https://github.com/danfickle/openhtmltopdf/issues/92#issuecomment-324269602


harbulot:
This seems to break CJK font rendering (see example in #129).

danfickle:
Hi @schmitch 

Thanks for the PR, but unfortunately we can’t eliminate non-Latin character set support, so I have to close this PR. Let’s continue the discussion in the PDF/A issue:

https://github.com/danfickle/openhtmltopdf/issues/92#issuecomment-331384337

Thanks also to @harbulot for testing this.

PR #132, opened 2017-09-21
check data:image type against revolved url

AlbanSeurat:
Using URI Resolver we can change the uri and therefore my guess is that the check on image should be on the URI resolved. 

For instance I have an uri with "cid" scheme which I transform into data:image as they are both kind of embedded images ...


AlbanSeurat:
@danfickle any chance to merge this PR ?

danfickle:
Hi @AlbanSeurat 

Thanks for the PR. I just have some outstanding code I need to commit before merging. I’ll try to work on it on Saturday.

danfickle:
Thanks again @AlbanSeurat 

I think this is theoretically a possibly breaking change but unlikely anybody has assumed order. I thought I would note it here, just in case. Merging now.

AlbanSeurat:
@danfickle thanks

AlbanSeurat:
Any change to have 0.0.1-RC12 to be release soon ?

danfickle:
@AlbanSeurat 

Just released RC-12 now.

AlbanSeurat:
Thanks

Alban - iPhone

PR #137, opened 2017-10-08
Fix #135: Don't write a MiterLimit of 0 into the PDF

### LINKED ISSUE #135: Dotted and dashed table border styles cause errors in Adobe Reader
Tables with cell borders styled dotted or dashed render fine in most pdf viewers, but generate an error or render as a blank page when viewer id Adobe Reader DC. Attached are two html files (one dotted, one dashed), and the pdf files they generate

[dashed-border-bug.html.txt](https://github.com/danfickle/openhtmltopdf/files/1342171/dashed-border-bug.html.txt)
[dashed-border-bug.pdf](https://github.com/danfickle/openhtmltopdf/files/1342168/dashed-border-bug.pdf)

[dotted-border-bug.html.txt](https://github.com/danfickle/openhtmltopdf/files/1342172/dotted-border-bug.html.txt)
[dotted-border-bug.pdf](https://github.com/danfickle/openhtmltopdf/files/1342169/dotted-border-bug.pdf)

rototor:
as Acrobat Reader does not like that and stops rendering everything ...

Also allow to only execute specific test files.


danfickle:
Thanks for tracking this down!

PR #141, opened 2017-10-10
Upgrade to PDFBoxGraphics2D 0.9. XForms are now finally compressed

rototor:
and images are reused.

As a side note: I would like to merge some of the changes made by @backslash47 https://github.com/backslash47/openhtmltopdf/

He made some nice commits with features and bugfixes. As far as I can see they are all fine. 

Only the PDF/A support API is not perfect yet. I would rather use a enum as parameter then a "magic" string constant. And map the enum internally to the string constant.

Should I make pullrequests for the single changes or could you cherrypick them?

### DIFF
--- a/openhtmltopdf-pdfbox/pom.xml
+++ b/openhtmltopdf-pdfbox/pom.xml
@@ -60,7 +60,7 @@
 	  <dependency>
 		  <groupId>de.rototor.pdfbox</groupId>
 		  <artifactId>graphics2d</artifactId>
-		  <version>0.8</version>
+		  <version>0.9</version>
 	  </dependency>
   </dependencies>
 


danfickle:
Wow, that’s some awesome work by @backslash47 

I think we can merge the lot, then change the builder to use a PDF/a enum.

Out of question, where does one get a color profile?

backslash47:
Hello,

the work around PDF/A is definitelly not done very well. I just needed a quick way to export PDF/A.

There are two possible problems:
1. It might be not generating a proper PDF/A. I did not have time to read the PDF/A spec, so I only filled the really necessary parts with data I've got. 
2. It is not very configurable (e.g. color profiles, embedding fonts, ...)

Other changes were done witch better care :) So use and/or change everything which is needed for the changes to be incorporated in the upstream. 

Best regards
Matus

rototor:
For PDF/A validation: PDFBox implements a validator, you can download and use it from the PDFBox website. If a generated document validates with this validator it _should_ be fine.  Otherwise it referes to the spec about what is violated.

rototor:
@danfickle Sorry, did not understand your question at first. To get a color profile in Java you usually have a .icc file for it. Or you use some of the default profiles.

I.e. ```ICC_Profile.getInstance(ColorSpace.CS_sRGB)```for the default sRGB profile or you load a profile from a input stream ```ICC_Profile.getInstance(inputStream)```.

What profile you use and want depends on the output you like to generate. CS_sRGB should be fine for most usecases. Advanced color profiles are usually needed for prepress, and you normally find them embedded into images - which is fine for PDF, and is hopefully fine for PDF/A. You can lookup  [here](https://github.com/rototor/pdfbox-graphics2d/blob/469618798174911537ea35590a2092982c33b5e4/src/main/java/de/rototor/pdfbox/graphics2d/PdfBoxGraphics2DLosslessImageEncoder.java#L73) how to embed a color profile into a PDF image.

Issue #128, opened 2017-09-14
Support for setting SVG dimensions via CSS


harbulot:
I'm trying to set the size of an SVG image using CSS styles (e.g.` style="width: 25%"`), but this is ignore.
Only the `width` and `height` attributes of the `<svg />` element seem to have an effect. The default values seem to be 400, but I'm not sure what unit this uses, or what it is in relation to the page width.

# Example

Here is a simple HTML example:

```html
<!DOCTYPE html>
<html>
<head>
<style type="text/css">
@page {
	size: A4 portrait;
	margin: 2cm 1cm;
}
</style>
</head>
<body>
    <h1>Test</h1>
    <h2>Text</h2>
    <div style="width: 25%; border: 1px solid black;">
        <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit.
            Donec venenatis velit enim, a placerat lectus viverra non.
            Proin varius porta ligula, in fringilla erat suscipit a.</p>
    </div>
    <h2>SVG</h2>
    <svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 30 20"
        style="width: 25%; border: 1px solid black;">
        <rect x="1" y="1" rx="5" ry="5" width="28" height="18"
            stroke="blue" fill="green" stroke-width="2" />
    </svg>
</body>
</html>
```

Here is the output in a browser:

![image](https://user-images.githubusercontent.com/80994/30446766-0093ef7c-9982-11e7-9259-e7274081be7a.png)

Here is the PDF output:

![image](https://user-images.githubusercontent.com/80994/30446928-7634f7e4-9982-11e7-9258-451fdf8c80ac.png)

Here is the code I've used:

```java
        try (InputStream is = getClass()
                .getResourceAsStream("/html/" + baseName + ".html");
                FileOutputStream os = new FileOutputStream(baseName + ".pdf")) {

            String docHtml = IOUtils.toString(is, StandardCharsets.UTF_8);
            PdfRendererBuilder builder = new PdfRendererBuilder();
            builder.useSVGDrawer(new BatikSVGDrawer());

            builder.withHtmlContent(docHtml, "");

            builder.toStream(os);

            builder.run();
        }
```

# Improvement attempt 1

I've looked into the SVG support code to investigate, and here is a partial fix:

```diff
--- a/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/PDFTranscoder.java
[...]

PR #144, opened 2017-10-18
Using Batik to compute SVG dimensions through hints.

danfickle:
Looks good. Thanks for this work.


danfickle:
See #128 for discussion.

PR #143, opened 2017-10-18
September improvements by @backslash47

danfickle:
Thanks for the awesome work @backslash47

PR #148, opened 2017-11-11
Fix #136: positive strikethrough offset for max comparision

### LINKED ISSUE #136: CSS { text-decoration: line-through; } doesn't work properly
CSS property { text-decoration: line-through; } (and "del" tag) doesn't work properly both in sandbox and in actual generated PDF. See picture
image

[test case](https://github.com/danfickle/openhtmltopdf/blob/eb69d4ecde6f886705bfcf747e56ae639078d396/tests/diff/fonts/text-decoration/all.xhtml)

alebar:
This PR will fix the #136.
Basically the offset should be negative but when the code looks for the biggest one, it should use a positive value, because Float.MIN_VALUE is always positive.


danfickle:
Thanks @alebar 

I have used your test case with the fix provided in #143 as it seemed simpler.

Thanks again.

Issue #54, opened 2016-11-30
Issue Running on JBoss EAP 6.1


alanhay:
I have an issue running on JBoss EAP 6.1 which seems to be down to some kind of Xalan / XML parser conflict. I note that the line in question has a //FIXME comment so wonder if you can advise of any workaround. My code is running fine on Jetty.

EAP stacktrace:

```
JBWEB000236: Servlet.service() for servlet dispatcher threw exception: java.lang.ClassNotFoundException: com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl from [Module "deployment.edental-web-pars.war:main" from Service Module Loader]
	at org.jboss.modules.ModuleClassLoader.findClass(ModuleClassLoader.java:213) [jboss-modules.jar:1.3.7.Final-redhat-1]
	at org.jboss.modules.ConcurrentClassLoader.performLoadClassUnchecked(ConcurrentClassLoader.java:459) [jboss-modules.jar:1.3.7.Final-redhat-1]
	at org.jboss.modules.ConcurrentClassLoader.performLoadClassChecked(ConcurrentClassLoader.java:408) [jboss-modules.jar:1.3.7.Final-redhat-1]
	at org.jboss.modules.ConcurrentClassLoader.performLoadClass(ConcurrentClassLoader.java:389) [jboss-modules.jar:1.3.7.Final-redhat-1]
	at org.jboss.modules.ConcurrentClassLoader.loadClass(ConcurrentClassLoader.java:134) [jboss-modules.jar:1.3.7.Final-redhat-1]
	at java.lang.Class.forName0(Native Method) [rt.jar:1.8.0_102]
	at java.lang.Class.forName(Class.java:348) [rt.jar:1.8.0_102]
	at javax.xml.transform.FactoryFinder.getProviderClass(FactoryFinder.java:116) [rt.jar:1.8.0_102]
	at javax.xml.transform.FactoryFinder.newInstance(FactoryFinder.java:169) [rt.jar:1.8.0_102]
	at javax.xml.transform.TransformerFactory.newInstance(TransformerFactory.java:152) [rt.jar:1.8.0_102]
	at com.openhtmltopdf.resource.XMLResource$XMLResourceBuilder.createXMLResource(XMLResource.java:216) [openhtmltopdf-core-0.0.1-RC8.jar:]
```

which is triggered by the following code:

```
try {
	// FIXME:
[...]

PR #155, opened 2017-11-30
Fixes #154

### LINKED ISSUE #154: Property 'http://javax.xml.XMLConstants/property/accessExternalDTD' is not recognized.'
Hi everybody, we are trying to use openhtml2pdf and we had some problems with xalan. After some bug we found where the exception has been thrown, and it was in 'openhtmltopdf-core/src/main;java/com/openhtmltopdf/resource/XMLResource.java' in line 222 aproximadetly, we just add a try/catch to ignore when the attribute XMLConstants.ACCESS_EXTERNAL_DTD is not supported by the xalan. Probably its not the best solucion for this problem, but its a workaround that solves the problem.

estevandiedrich:


danfickle:
Thanks @estevandiedrich 

I have manually incorporated your changes together with allowing the user to manually configure which TransformerFactory to use, so I'll close this PR. We can continue discussion on #54

Issue #38, opened 2016-09-27
Implement CSS3 transform


rototor:
I've started implementing CSS3 transform. My primary goal is to get rotate() working. The rest should be relative easy, but I primary need rotate(90deg).

You can view the my current development here https://github.com/rototor/openhtmltopdf/tree/css3-transform-implementation

CSS parsing should be mostly done, but it is not yet applied on the elements. I'll be away for a week on holidays, but will continue to implement this afterwards.

Note, related to #23 I implemented saveSate() and restoreState() - also for Java2D.

danfickle:
@rototor - Looks good. The hard part is implementing transform origin I think (from memory).

danfickle:
@rototor - Would you mind submitting a pull request with what you have got and I'll see if I can help with implementation?

Thanks,
Daniel.

P.S I hope you had a wonderful holiday!

rototor:
@danfickle I sent you a pull request #45. Sorry for the delay, after holiday the usual "we need that and that now" happened in work... had no time yet to look future into this...

danfickle:
Thanks mate, I'll see how far I get.

danfickle:
OK, initial support has landed.

What is not supported yet:
+ Multiple transforms on the one element.
+ Transformed element across pages.
+ ````translate, translateX, translateY, skew, skewX and skewY````

What is supported:
+ ````rotate, matrix, scale, scaleX, scaleY````
+ Transforms on nested elements.
+ transform-origin.
+ Transforms on the second and subsequent page.

I'll do a release tomorrow so everyone can test it out.

danfickle:
I thought about this some more, the problem with cross-page transforms is that the page(s) to paint text and other content on is determined by the pre-transform coordinates. We need to use the post transform coordinates.

danfickle:
All working now.

PR #157, opened 2017-12-14
#38 Fix css3 transform

rototor:
The current CSS3 transform support is still broken. In this branch I was ably to fix it mostly for Java2D. But nothing has changed with the PDF output.

I've created a sample file to test this. The sample file renders this in Safari:

image

This way using Java2D:
image
You may note that in the second line (scaled and rotated) the resulting div for 180 degree and 45 degree has some strange offset, but everything else is fine.

But it is still broken when generating a PDF:
image

@danfickle I tried around all different things, but I currently have no idea what offset is applied wrong here for the PDF output. Could you please take a look into that?

I did add some generic parameters to some lists. I hope you don't mind.

Thanks.
P.S.: Can we just remove the old Java2DOutputDevice stuff? i.e. com.openhtmltopdf.swing.Java2DRenderer and com.openhtmltopdf.swing.Java2DOutputDevice?

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/css/parser/CSSParser.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/css/parser/CSSParser.java
@@ -19,38 +19,23 @@
  */
 package com.openhtmltopdf.css.parser;
 
-import java.io.IOException;
-import java.io.Reader;
-import java.io.StringReader;
-import java.net.URI;
-import java.net.URISyntaxException;
-import java.net.MalformedURLException;
-import java.net.URL;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.logging.Level;
-
-import org.w3c.dom.css.CSSPrimitiveValue;
-
 import com.openhtmltopdf.css.constants.CSSName;
 import com.openhtmltopdf.css.constants.MarginBoxName;
 import com.openhtmltopdf.css.extend.TreeResolver;
 import com.openhtmltopdf.css.newmatch.Selector;
 import com.openhtmltopdf.css.parser.property.PropertyBuilder;
-import com.openhtmltopdf.css.sheet.FontFaceRule;
-import com.openhtmltopdf.css.sheet.MediaRule;
-import com.openhtmltopdf.css.sheet.PageRule;
-import com.openhtmltopdf.css.sheet.PropertyDeclaration;
-import com.openhtmltopdf.css.sheet.Ruleset;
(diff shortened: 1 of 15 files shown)


danfickle:
I see you're still adding commits so I'll hold off on merging. I did however find out why my old test cases were working in PDF but not your examples. It turns out the body/page margin was breaking things. Try the following in the sandbox and you get correct results. As soon as a body/page margin is added however, it breaks.
````html
<html>
<head>
	<style>
		@page { margin: 0; }
		body { margin: 0; }
		div.block > div {
			border: 1px solid blue;
			display: inline-block;
			padding:5px;
		}
		div.block {
			width: 100%;
			border: 1px solid green;
			margin-bottom:3cm;
		}
	</style>
</head>
<body>

<h1>Sample to demonstrate CSS Transform</h1>

<div class="block">
	<div> 0</div>
	<div style="transform:rotate(90deg)"> 90</div>
	<div style="transform:rotate(180deg)"> 180</div>
	<div style="transform:rotate(270deg)"> 270</div>
	<div style="transform:rotate(45deg)"> 45</div>
	<div style="transform:rotate(135deg)"> 135</div>
</div>
</body>
</html>
````

The other problem is the box collector is not transform aware. The box collector is responsible for collecting lists of boxes for the current page. So if a rotate, for example, causes a box to move over a previous or following page, it won't show on the extra page.

I did have a try at fixing the box collector without much luck a couple of weeks ago. But thinking about it, I think we will need an absolute transform that is in effect for every layer to get the box collector to work.

danfickle:
By the way, feel free to delete/move Java2D related output device code out of the core.

rototor:
@danfickle I had some spare minutes, so I tried around a little bit. But I will go to dinner now and also won't have time till evening to look future (here it is 11:25 AM).

So fill free to work in this branch. I wont touch it till evening.

rototor:
@danfickle Thanks for the hint. I fixed it for the PDF case. Did not test if transform works correctly in the margin area. But for my use case thats enough.

The detection if the margin must be considered is a little bit dirty at the moment, as I use the flip factor to detect if this is a PDF output. But at least now it works for both the PDF and the PNG output case.

rototor:
@danfickle I'm going to do some future work in this branch. So you may want to wait with the merge. Especially I want to

- cleanup the "flip" flag.
- get transform working in the page borders.
- maybe even get my hands dirty with translate() - depending how much time I have later today.

rototor:
@danfickle  And I think that is fully ok for a transformed element to be not correctly split over multiple pages. I don't think someone would really want this. A nice looking report should not have some  "cut  in the middle by page break" elements in it... This would look broken IMHO.

rototor:
@danfickle I was not able to do as much today as I hoped. Would you mind merging the current state? Top/Left/TopLeft-Corner margins work correctly with transforms, the other margins don't work yet with transforms.

PR #158, opened 2017-12-15
make it possible to set a producer

schmitch:
currently the producer defaults to `openhtmltopdf.com` while I think that is sane in the most cases. it's sometimes not wished to actually "show" which library was used to produce pdfs.


rototor (approved):
I don't like that the constructor arguments are going to explode as their are more and more arguments.

But other than that, this looks fine to me.

danfickle:
Thanks again @schmitch 
Thanks for taking a look @rototor 

Merging now.

PR #162, opened 2017-12-22
Old java2d cleanup

rototor:
This mostly deletes old stuff.  Tons of old stuff.

But I found the old image-map parser code, which I tried to refactor to reuse it for the PDF. That is not really working correctly yet, as in PDF you can specify shapes using quads. Need to first lookup how to correctly convert the polyline paths to quads ...

When I get this working you can specify using "usemap" a image-map for img, svg and object. This could come handy for some diagrams.

Should I split this merge request? I.e. seperating the deletion work from the image-map stuff?


danfickle:
Hi @rototor 

This cleanup is much needed. Image map support will be a great bonus. 

Thanks and have a wonderful Christmas.

PR #163, opened 2017-12-26
Get ImageMap Shape tessilation right

rototor:
Using https://www.sunshine2k.de/coding/java/Polygon/Kong/Kong.html 

I've adapted the class for this use case. In the readme provided with the java sources the author says "please give me some credits". IANAL but I think it should be save to include the files here. But I will check back with the author. 

From the given transform only the scale is used, as I could not get the concatination of the transform working correctly. This should only affect images affected by CSS-transform. Feel free to fix this, to make this work correctly with CSS-transforms.

In the sample svg-inline there is now a image map using two rectangles and a circle. Those work with Acrobat Reader with this changes. Mac OS X preview seems to ignore the quad points.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ head - 0.0.1-RC13-SNAPSHOT
 ========
 + [Fix for XML loading problems with JBoss,Wildfly](https://github.com/danfickle/openhtmltopdf/issues/54) Thanks @estevandiedrich, @alanhay
 + [Support image maps on img, object and svg](https://github.com/danfickle/openhtmltopdf/pull/163) Uses a kong polygon tesselation implementation by [sunshine2k](https://www.sunshine2k.de/coding/java/Polygon/Kong/Kong.html)
++ BINARY API BREAKING CHANGE: [Support custom shape <=> link maps in object drawers](https://github.com/danfickle/openhtmltopdf/pull/163)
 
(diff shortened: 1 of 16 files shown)


rototor:
Just got the information from the author of the kong algorithm implementation that the source is licensed using MIT and that we can freely use it. We can mention him in the release notes, but we don't have to.

I'll add the changes for that and also will fix the transform / scaling stuff. So you may wait a day or two until you merge this branch, as I will future work on it.

rototor:
@danfickle I'm done with my changes here. Currently the transforms applied through CSS "transform" are not applied on the transform which is given to the link manager. 

Because of this transformed objects have their links wrong:

image

Other than that it seems ok now. This PullRequest should now be ready for merge.

Some other thing: To test this out I've integrated JFreeChart into the examples and build some small example object drawer which allow to define graphs using objects in HTML. Should we provide such object drawers in e.g. a openhtmltopdf-objects library? Then openhtmltopdf could draw different kinds of diagrams "out of the box". What do you think?

danfickle:
Thanks @rototor - I have no problem with extra modules to incorporate different object drawers. It seems like a pretty good idea. The only caution I have is documenting all the new feature we’re (well, mostly you - thank you) adding.

PR #165, opened 2018-01-04
WIP: Make a basic Objects Package

rototor:
@danfickle  This is not yet meant to be merged. I just got a problem where I could use your help.

See here:
[featuredocumentation.pdf](https://github.com/danfickle/openhtmltopdf/files/1604354/featuredocumentation.pdf)

It tries to add a background-PDF by placing a special object into topleft corner. But for what ever reason the object in the topleft corner is only painted on the first page. No idea why...

To run this just execute the FreeMarkerGeneratorTest.

Any idea what could be causing this?


rototor:
I've got the problem fixed with the pdf-background. The documentation needs future work but it's at least a start. 
[featuredocumentation.pdf](https://github.com/danfickle/openhtmltopdf/files/1720702/featuredocumentation.pdf)

The pdf-background feature allows to implement watermarks etc. using a PDF document. I.e. to apply CI backgrounds etc. 

This PDF is written into openhtmltopdf-examples/target/tetst/freemarker directory by the FreeMarkerGeneratorTest.

It still shows some problems e.g. the rotated "FEATURE DOCUMENTATION" on the left side is only on the first page. I assume the that transform in the margin area is still broken on multiple pages. Or maybe it may get fixed by #171 - i did not test that yet. (i.e. I did not merge both branches together).

PR #171, opened 2018-02-02
Testcase and fix for rotation on the second and following pages.

rototor:
We must add the previous vertical page margins.

The test case is also an example for a more complicated report table.

Also renamed setPdfMatrix to applyPdfMatrix, because it's not a replacement of the matrix but rather a addition to the current matrix. And we can spare the applying of the reverse matrix in the PDF case, as we can just saveGraphics()/restoreGraphics().

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/Layer.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/Layer.java
@@ -429,6 +429,16 @@ public class Layer {
 			RectPropertySet margin = c.getPage().getMargin(c);
 			relTranslateX += margin.left();
 			relTranslateY += margin.top();
+			
+			/*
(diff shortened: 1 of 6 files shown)


danfickle:
G’day @rototor 

I’m not sure we can just use save/restore graphics as it is also in use for clip/setClip. The problem is that these may be out of order, for example:
Clip (save)
Transform (save)
SetClip (restore wrong!)
Pop transforms (restore)

This was why I came up with the convoluted system of inverse transforms and reapplying transforms in effect after a setClip call. Ideally, we could guarantee that clip/setClip and pushtransforms/poptransforms are only called in correct order, but I found the clipping system to be complex enough to make this difficult. Incidentally, save/restore is the only way to implement setClip.

rototor:
@danfickle Ok, I restored the old behavior and removed the call to save/restore from the transform stuff.

danfickle:
Thanks @rototor 

Looks good.

Issue #161, opened 2017-12-21
MathML Support


m-a-t:
MathML seems to be the technology how to embed formulas in web pages.

Consider adding support in openhtmltopdf.

There is a project based on flyingsaucer which adds this support: https://mvnrepository.com/artifact/com.github.rjolly/flying-saucer/9.1.1
(However, GPL licensed) Maybe this rjolly can contribute it to openhtmltopdf

rjolly:
Hi mat,

I would be pleased to do so, however there are some caveats. Regarding the license, it is not an issue. I choose GPL because my fork is mainly a browser and not a library. But I do not even know if I have the right to do this, as flyingsaucer is LGPL. So do not hesitate do borrow anything from my project. The commit relevant to MathML addition is https://github.com/rjolly/flying-saucer/commit/f4c4a9d74e521b82e2a498be98387d4321cc24cd

However, at the moment, it works only for screen rendering, and not PDF. The reason is, that I am using JEuclid, which itself uses FOP, whereas flying-saucer uses iText. So having it work in openhtmltopdf/PDF-Box is a little more work than it seems I think.

rototor:
@rjolly Only having "screen" rendering, i.e. drawing to a Graphics2D is perfectly fine. This is what  [PDFBox-Graphics2D](https://github.com/rototor/pdfbox-graphics2d) is for. Even integrating JEuclid into openhtmltopdf for a quick test is easy using object drawers #78:

```kotlin
        objectDrawerFactory.registerDrawer("custom/mathml", { e, x, y, width, height, outputDevice, ctx, dotsPerPixel
            ->
            val dummy: CustomEvent? = null
            var src = ""
            for (i in 0..e.childNodes.length) {
                val item = e.childNodes.item(i)
                if (item is org.w3c.dom.CharacterData)
                    src = item.data
            }
            if (src.startsWith("<["))
                src = src.substring(2)
            val node = net.sourceforge.jeuclid.parser.Parser.getInstance().parse(StreamSource(StringReader(src)))
            val jeuclidDom = DOMBuilder.getInstance().createJeuclidDom(node)
            val realWidth = width / dotsPerPixel
            val realHeight = height / dotsPerPixel
            outputDevice.drawWithGraphics(x.toFloat(), y.toFloat(), realWidth.toFloat(), realHeight.toFloat(), { gfx ->
                /*
                 * Scale
                 */
                val viewer = jeuclidDom.defaultView
[...]

PR #177, opened 2018-02-15
Latex support

rototor:
This is an addition to #161 - now also basic LaTeX formulars are supported. I've updated the feature documentation a little bit [featuredocumentation.pdf](https://github.com/danfickle/openhtmltopdf/files/1729635/featuredocumentation.pdf).

Also the builder share now a common base class as many methods are just copy&paste as they mostly share the same settings.

To support LaTeX SnuggleTex is used. To do so I've created a FSDOMMutator interface, which can register callbacks to modify the W3C-Document before it is evaluated as HTML. For LaTeX the <latex> element text is converted into DOM nodes which are then appended as child. The real math rendering then happens using the MathML support.

I also added to the featuredocumentation the generation of a Table Of Content. But prepending this ToC to the document using content:element(toc) in an :after element seems not to work. But maybe I find another solution.

As always I you have better names for some classes, feel free to request changes.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/context/StyleReference.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/context/StyleReference.java
@@ -27,6 +27,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.logging.Level;
 
+import com.openhtmltopdf.css.sheet.FontFaceRule;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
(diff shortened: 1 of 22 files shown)


danfickle:
@rototor 

This is excellent. My only query is adding the svg-support module as a dependency of the mathml module. Was this intended? Anyway, I’ll merge now and we can remove if unintentional.

rototor:
@danfickle MathML needs Batik (indirect through JEuclid). If there is a new Batik release, all modules using Batik should use the same version of Batik. The svg-support module specifies a Batik version, so I added it as dependency to "force" that Batik version on the other modules.

But using explicit dependency management in the parent pom.xml makes more sense. 

=> You'll get a pull request for that.

Issue #187, opened 2018-03-27
Possible ClassCastException when createXMLResource() uses xerces


sosnut:
When xerces is used in the application, createXMLResource() can send the following exception:
_org.apache.xerces.parsers.XIncludeAwareParserConfiguration cannot be cast to org.apache.xerces.xni.parser.XMLParserConfiguration_

```
java.lang.ClassCastException: org.apache.xerces.parsers.XIncludeAwareParserConfiguration cannot be cast to org.apache.xerces.xni.parser.XMLParserConfiguration
	at org.apache.xerces.parsers.DOMParser.<init>(Unknown Source)
	at org.apache.xerces.parsers.DOMParser.<init>(Unknown Source)
	at org.apache.xerces.jaxp.DocumentBuilderImpl.<init>(Unknown Source)
	at org.apache.xerces.jaxp.DocumentBuilderImpl.<init>(Unknown Source)
	at org.apache.xerces.jaxp.DocumentBuilderFactoryImpl.setFeature(Unknown Source)
	at com.openhtmltopdf.resource.XMLResource$XMLResourceBuilder.createXMLResource(XMLResource.java:200)
	at com.openhtmltopdf.resource.XMLResource$XMLResourceBuilder.access$100(XMLResource.java:164)
	at com.openhtmltopdf.resource.XMLResource.load(XMLResource.java:77)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.setDocumentFromStringP(PdfBoxRenderer.java:349)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.<init>(PdfBoxRenderer.java:255)
	at com.openhtmltopdf.pdfboxout.PdfRendererBuilder.buildPdfRenderer(PdfRendererBuilder.java:106)
	at com.openhtmltopdf.pdfboxout.PdfRendererBuilder.run(PdfRendererBuilder.java:84)
```
(Observed with Oracle Jdk 1.8.0_66, openhtmltopdf 0.0.1-RC12).

By default openhtmltopdf does not use xerces (but in some cases it seems to pick xerces implementation of `DocumentBuilderFactory` if it's used elsewhere in the application; I suppose it would be better to use the internal JAXP implementation from _com.sun.org.apache.xerces.internal.jaxp_ unless specified otherwise – since xerces is quite old now).

This issue seems similar to issue #54 (except that Xerces is involved instead of Xalan).

[...]

PR #188, opened 2018-03-28
#187 Specifying default DocumentBuilderFactory implementation and add…

sosnut:
…ing a way to bypass it

Specifying "com.sun.org.apache.xerces.internal.jaxp.DocumentBuilderFactoryImpl" as the default DocumentBuilderFactory, in BaseRendererBuilder. (This is the factory that is used by default, but an application may depend on an other parser which could provide an incompatible builder factory).

Adding method useDocumentBuilderFactoryImplementationClass(), to be able to use another implementation (or the DocumentBuilderFactory currently available in the application, if null is passed) if necessary.


sosnut:
Changes very similar to fix of #54 in https://github.com/danfickle/openhtmltopdf/commit/a737e965fe3032e3967bbf88142326e99317d958

danfickle:
Thanks @sosnut 

This is excellent. Merging now.

sosnut:
Great; thanks to you for the merge.

PR #193, opened 2018-04-06
Upgrade the PDFBox Version to 2.0.9

rototor:
@danfickle I think you wanted to upgrade the version for -RC13 but just simply forgot it :)


danfickle:
Thanks @rototor 

However, shouldn’t we update pdfbox-graphics2d at the same time to avoid PDFBOX version hell?

rototor:
Yes, you are right, i release a updated version.

Issue #196, opened 2018-04-08
custom UserAgent / PdfBoxUserAgent


delafer:
I am Alex and I'm trying to move my project to openhtmltopdf & openbox instead of using itext | fop based flyingsaurcer.
 

- **Сoncretely:**

I don't see any handsomely set & use own custom **UserAgent / PdfBoxUserAgent** , like I have this already done in flyingsaurcer in oder to load some resources (images, css) directly from classpath. Of course, I can use URLStreamHandlerFactory, but It's tricky and not always possible in AS like BEA/Oracle WL.

- **Little philosophy & critics:**

I see a lot of design problems in this library. IMHO architecture of this library comparitevly to flyingsaurcer/neoFlyingSaurcer has not improved, but sometimes was even worsened. Constructors with >= 18 parameters, huge methods & consturctor intitializers, sometimes unreasoned method & fields modifiers, its looks like procedural  instead of object-oriented programming,because of lack of factories, I don't even say about other patterns like decorators, adapters, etc. -> to make the code cleaner & expendable.
It's works of course, while walking straight and not trying to do something special, something except "hello world" ;-)
 

- **Finally my praise and compliments:**

Your working hardly, I see a lot of pull/push request in git, I like SVG support (_and it works, BTW: I have done something similar with flyingsaurcer using batik & ReplacementElementFactory_), I like the rate of new versions release.
Idea behind flyingsaurcer was genial, for reports generation I have used before Jasper, BIRT, and pure iText libraries.
I thank you for your work and with use a good luck.
 
_P.S. think a little bit not only about a new features, but also about design and extensibility of openhtmltopdf. About clean and well-structured code.
I can do a pull request to git to allow use/setting of custom PdfBoxUserAgent, but I am not sure, that will you accept this / third party code / changes._

Kindly regards,
Alexandr

![example](https://user-images.githubusercontent.com/5951449/38465731-dabe3eac-3b1f-11e8-90ab-b935a0b11669.png)

This code doesn't work as expected, because 
**builder.buildPdfRenderer()** instantiates ->  **new PdfBoxRenderer(...)** instance and during instantiation in constructor **new PdfBoxUserAgent()** instance will be created and in the same place in constructor document will be parsed and method private void **setDocumentP(Document doc, String url, NamespaceHandler nsh)** will be called during initialization and following line in this method ->
[...]

Issue #170, opened 2018-01-16
RFC - Roadmap for version 1


danfickle:
These are my thought on the issues that need to be addressed before version 1 is released, in no particular order:
+ [x] #161 - MathML support - COMPLETE
+ [x] NO-ISSUE-YET, COMPLETE - Entity support such as nbsp for HTML, InvisibleTimes for MathML and SVG entities. This is tricky as there is [no way to programatically inject entities using the Java XML parser](https://stackoverflow.com/questions/36026353/parsing-xml-file-containing-html-entities-in-java-without-changing-the-xml).  <s>I propose that we add a doctype dynamically to the start of the XML input, with the desired entities. However this means we have to read XML input into a string, rather than just passing a file or input stream to the XML reader. The builder can be used to specify which entities to load.</s> Used custom doc types and an external entity resolver instead. DOCUMENT.
+ [ ] #38 - Transforms. A few issues remaining to implement:
   + [x] Link placing doesn't take account of transforms.
   + [x] Translate is not implemented.
   + [x] Some work for transformed boxes in page margins.
   + [ ] Testing. Do transforms of MathML, SVG and custom objects?
+ [x] NO-ISSUE-YET - Logging / error handling overhaul - Currently error handling is ad-hoc. For example should we continue on a load failure or fatally throw? I propose to allow this to be configurable by allowing the user to hook logging on a per-run basis and halt on any log message (which will be changed to enum constants) with a poison exception.
+ [x] #60 - CSS3 Columns - Currently implemented for text only. Need to debug to allow other box types in columns.
+ [x] #126 - Overflowing pages - Currently content that goes past the right margin is cut off silently. This is mostly a problem with tables. I propose a CSS property that allows cut off content to be printed on the next page. DOCUMENT.
+ [x] #204 - Multi run cache - Currently there is a multi-run cache hook method, but the objects stored may not be thread safe. This means it is unsuitable for many use-cases. Propose to remove all caches except font metrics cache.
+ [ ] NO-ISSUE-YET - Per run cache - Need to make sure nothing is being placed into a PDF document more than once. For example, is an img from the img tag and a background image from the same url embedded twice?
+ [x] #83 - Unicode font justification fix - There is a fix in #143 but we are waiting for PDF-BOX 2.0.9 to implement it.
+ [ ] #123 - RTL table layout - Altering table layout to correct RTL scares me but there have been a couple of requests so should try.
[...]

PR #198, opened 2018-04-08
for #196 - custom UserAgent / PdfBoxUserAgent

delafer:
Factory to set custom PdfBoxUserAgents

### DIFF
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
@@ -1,0 +1,95 @@
+package com.openhtmltopdf.pdfboxout;
+
+/**
+ *  PdfBoxUserAgentFactory. Singleton.
+ */
+public class PdfBoxUserAgentFactory {
+
+    private static class LazyFactoryHolder {
+        private static final PdfBoxUserAgentFactory INSTANCE = new PdfBoxUserAgentFactory();
+    }
+
+    protected static PdfBoxUserAgentFactory instance() {
+        return LazyFactoryHolder.INSTANCE;
+    }
+
+    private static class LazyHolder {
+        private static final UserAgentFactory INSTANCE = new UserAgentFactory() {
+            public PdfBoxUserAgent getPdfBoxUserAgent(PdfBoxOutputDevice outputDevice) {
+                return new PdfBoxUserAgent(outputDevice);
+            }
+        };
+    }
+
+    /*
+     * In all versions of Java, the idiom enables a safe, highly concurrent lazy initialization with good performance.
+     */
+    private static UserAgentFactory factoryDef() {
+        return LazyHolder.INSTANCE;
+    }
+
+    private ThreadLocal<UserAgentFactory> threadLocal;
+    private UserAgentFactory global;
+
+
+    {
+        threadLocal = new ThreadLocal<UserAgentFactory>();
+    }
+
+    /**
+     * Register [global] own / custom PdfBoxUserAgentFactory
+     * @param factory - implementation of UserAgentFactory interface
+     */
+    public void registerGlobalPdfBoxUserAgentFactory(UserAgentFactory factory) {
+        this.global = factory;
+    }
+
+    /**
+     * UnRegister [global] own / custom PdfBoxUserAgentFactory
+     */
+    public void unregisterGlobalPdfBoxUserAgentFactory() {
+        this.global = null;
+    }
+
+    /**
+     * Register [threadLocal] own / custom PdfBoxUserAgentFactory
+     * @param factory - implementation of UserAgentFactory interface
+     */
+    public  void registerPdfBoxUserAgentFactory(UserAgentFactory factory) {
+        threadLocal.set(factory);
+    }
+
+    /**
+     * UnRegister [threadLocal] own / custom PdfBoxUserAgentFactory
+     */
+    public  void unregisterPdfBoxUserAgentFactory() {
+        threadLocal.remove();
+    }
+
+    /**
+     *
+     * @param outputDevice - instance of PdfBoxOutputDevice, mandatory field
+     * @return instanceof PdfBoxUserAgent
+     */
+    public static PdfBoxUserAgent getPdfBoxUserAgent(PdfBoxOutputDevice outputDevice) {
+        return instance().getPdfBoxUserAgent0(outputDevice);
+    }
+
+
+    private PdfBoxUserAgent getPdfBoxUserAgent0(PdfBoxOutputDevice outputDevice) {
+        UserAgentFactory thLocal = threadLocal.get();
+        return (thLocal != null ? thLocal : global != null ? global : factoryDef()).getPdfBoxUserAgent(outputDevice);
+    }
+
+
+    public interface UserAgentFactory {
+        /**
+         * Factory to use custom implementation of PdfBoxUserAgent classes
+         * @param outputDevice
+         * @return
+         */
+        public PdfBoxUserAgent getPdfBoxUserAgent(PdfBoxOutputDevice outputDevice);
+    }
+
+
+}
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfRendererBuilder.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfRendererBuilder.java
@@ -42,6 +42,15 @@ public class PdfRendererBuilder extends BaseRendererBuilder<PdfRendererBuilder> 
 	}
 
 	/**
+	 * Returns an instance of PdfBoxUserAgentFactory in case you<br/>
+	 * need to use custom implementation of {@link PdfBoxUserAgent} class
+	 * @return PdfBoxUserAgentFactory instance
+	 */
+	public PdfBoxUserAgentFactory getPdfBoxUserAgentFactory() {
+		return PdfBoxUserAgentFactory.instance();
+	}
+
+	/**
 	 * Build a PdfBoxRenderer for further customization. Remember to call
 	 * {@link PdfBoxRenderer#cleanup()} after use.
 	 *
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxRenderer.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxRenderer.java
@@ -127,7 +127,7 @@ public class PdfBoxRenderer implements Closeable {
         _outputDevice = new PdfBoxOutputDevice(DEFAULT_DOTS_PER_POINT, testMode);
         _outputDevice.setWriter(_pdfDoc);
         
-        PdfBoxUserAgent userAgent = new PdfBoxUserAgent(_outputDevice);
+        PdfBoxUserAgent userAgent = PdfBoxUserAgentFactory.getPdfBoxUserAgent(_outputDevice);
 
         if (httpStreamFactory != null) {
             userAgent.setHttpStreamFactory(httpStreamFactory);


openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
```diff
@@ -0,0 +1,95 @@
+package com.openhtmltopdf.pdfboxout;
+
+/**
+ *  PdfBoxUserAgentFactory. Singleton.
+ */
```
rototor, line 5: Oh no! Please no singletons....

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
```diff
@@ -0,0 +1,95 @@
...
+    }
+
+    /*
+     * In all versions of Java, the idiom enables a safe, highly concurrent lazy initialization with good performance.
+     */
+    private static UserAgentFactory factoryDef() {
+        return LazyHolder.INSTANCE;
+    }
+
+    private ThreadLocal<UserAgentFactory> threadLocal;
```
rototor, line 31: This smells like memory / class loader leaks in Tomcat etc. on WAR unloading... ThreadLocals are hard to do right in a servlet context.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
```diff
@@ -0,0 +1,95 @@
...
+     */
+    public  void registerPdfBoxUserAgentFactory(UserAgentFactory factory) {
+        threadLocal.set(factory);
+    }
+
+    /**
+     * UnRegister [threadLocal] own / custom PdfBoxUserAgentFactory
+     */
+    public  void unregisterPdfBoxUserAgentFactory() {
+        threadLocal.remove();
```
rototor, line 66: If you forget this, you will have a class loader leak in your servlet container on unload....

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfRendererBuilder.java
```diff
@@ -41,6 +41,15 @@ public void run() throws Exception {
 		}
 	}
 
+	/**
+	 * Returns an instance of PdfBoxUserAgentFactory in case you<br/>
+	 * need to use custom implementation of {@link PdfBoxUserAgent} class
+	 * @return PdfBoxUserAgentFactory instance
+	 */
+	public PdfBoxUserAgentFactory getPdfBoxUserAgentFactory() {
+		return PdfBoxUserAgentFactory.instance();
```
rototor, line 50: Thats bad API design. It also breaks the builder contract (i.e. everything but build() *always* returns PdfRendererBuilder for method chaining). Better would be 

- Just define the UserAgentFactory interface (as above)
- The builder should have a field userAgentFactory with a default implementation assign i.e.
```
        _userAgentFactory = new UserAgentFactory() {
            public PdfBoxUserAgent getPdfBoxUserAgent(PdfBoxOutputDevice outputDevice) {
                return new PdfBoxUserAgent(outputDevice);
            }
        };
```
- A setter to override the default implementation. i.e.
```
public PdfRendererBuilder setUserAgentFactory(UserAgentFactory customFactory) {
  this._userAgentFactory = customFactory;
  return this;
}
```
Of course it would require an additional parameter to the PdfBoxRenderer....

rototor (changes requested):
Thanks for the PullRequest, but the PdfBoUserAgentFactory is a bad design. No singletons please. Sorry.

Singletons with thread locals just cry for class loader leaks in servlet environments. Thats really hard to get right. There is a reason why your application code should just use guice, spring or maybe even dagger for singletons (and all other injections of course). Singletons should only be ```@Singleton``` annotated and injected classes.

Also this breaks the API contract of the builder and does not fit the current design. As the builder should only local *store* settings for the PdfBoxRenderer, but not expose global all builder affecting objects....

It's nice that you try to improve the library, and I am really sorry that I must criticize you here.

rototor:
FYI: If you want to get a guess what all strange kind of problems you get with singletons and ThreadLocals just read the documentation about this in the Finalzer class of Guava:

https://github.com/google/guava/blob/master/guava/src/com/google/common/base/internal/Finalizer.java

rototor:
Hmm, the more I think about this, the more I don't like it to let the user override the UserAgent. In FlyingSourcer there was no clean API to do things, you were more or less forced to directly override internals of the library. There were even multiple ways to do the same thing.

In OpenHTMLToPDF we try to provide a clean API (namely the *Builders and *Renderers) to customize all stuff you need to customize. The UserAgent is way to internal to be overridden by the user. If you need some functionality which would require you in FlyingSourcer to override the UserAgent we should  provide a clean way with the Builder.

In terms of Java 9+ modules: Only ```PdfRendererBuilder``` and ```PdfBoxRenderer``` should be accessible, same for the Java2D-classes. Of course, all interfaces  etc you can give to the Builders like the FSObjectDrawerFactory, FSDOMMutator and so on must be public and accessible too. Everything else is internal and should not be accessible to outside of the OpenHTMLToPDF library itself. Also openhtmltopdf-core should mostly only be accessible by openhtmltopdf-pdfbox and openhtmltopdf-java2d (in Java 9+ modules speak, with Java 8 you can not really restrict this beside using the very complicated OSGi).

Putting such hard restrictions on the accessibility of library internals would allow us to later refactor many internals without breaking user code. What do you think about this @danfickle?

For this reason I NAK the overall idea of this pull request.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
delafer: Singleton & Multiton patterns appeared a long before Guice, Spring iOC, Weld, Dagger, etc. 
 "Singletons should only be @Singleton annotated and injected classes."
who has said such a stupid thing? 
Singletons should only be @Singleton annotated in CDI / DI Enviroments, where dependency injection container is used, without CDI container, you can still use and write Singletons and where are normal classes which have static fields-like behavior.

"does not fit the current design"
LOL. I do not see any design at all. (PdfBoxRenderer = 18 args constructor + 108 lines of code in constructor init method. wonderful design. I am working for about 20 years as developer and have rarely seen such a wonderful design)

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
delafer: I have never had any problems with  ThreadLocal clases either Tomcat, Jetty, OC4J, Wildfly, Weblogic, etc. I can list / enumerate dozens of well-known / opensource projects using this standard Java class. It's available since JAVA 1.2 ;-) ThreadLocal could lead to memory leaks, but only if you don't know how to use it.

If only this library (openhtmltopdf) was developed using a thoughtful design it would not be necessary at all to create something like this. I almost threw up as I have seen constructor with 18 arguments and constructor itself has 108 lines of code! schoolchildren writing better code.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
delafer: just like with streams, if you opened it - you must close it. It is not an argument not to use them at all. Tomcat trying to reuse user threads, which could lead to problems with threadlocal ,what why tomcat developers recommends to clean them manually. Anyway tomcat was never a good choice for a production use in a serious projects.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfRendererBuilder.java
delafer: I have a suspicion that you just did not understand the code above. 
"build() always returns PdfRendererBuilder for method chaining"
Of course it returns always a PdfRendererBuilder, it should be used in this manner:

        PdfBoxRenderer renderer = null;
        try {
            builder.getPdfBoxUserAgentFactory().registerGlobalPdfBoxUserAgentFactory(new PdfBoxUserAgentFactory.UserAgentFactory() {
                @Override
                public PdfBoxUserAgent getPdfBoxUserAgent(PdfBoxOutputDevice pdfBoxOutputDevice) {
                    return new ExtendedPdfBoxUserAgent(pdfBoxOutputDevice);
                }
            });
            renderer = builder.buildPdfRenderer();
            //customUserAgent(renderer);
            renderer.layout();
            renderer.createPDF();
        } finally {
            if (renderer != null)
                renderer.close();
        }

delafer:
I think that you have sharply reacted to my critic. And that's why they responded reciprocally, beginning to criticize my proposal. This is unproductive. I assumed that my pull-request would be rejected. Therefore, I'm not surprised. Programmers do not like criticism. Most people believe that they write nice well-designed and thoughtful code.  I don't want to debate who "has bigger balls". Of course you, as a owners of the project. in theory this open-source project, in real life this project is developed by 2 people with a very strange ideas about software / library design.

delafer:
> Hmm, the more I think about this, the more I don't like it to let the user override the UserAgent. In FlyingSourcer there was no clean API to do things, you were more or less forced to directly override internals of the library. There were even multiple ways to do the same thing.
> 
> In OpenHTMLToPDF we try to provide a clean API (namely the *Builders and *Renderers) to customize all stuff you need to customize. The UserAgent is way to internal to be overridden by the user. If you need some functionality which would require you in FlyingSourcer to override the UserAgent we should provide a clean way with the Builder.

if the resource in the html template does not starts with the protocol name and instead of http://www..xyz.jpg defined not a absolute, but relative path e.G. url="/xyz/myImage.jpg" new library throws NPE exception, flyingsaurcer works without exceptions. this is only one example of a dozen others, where openhtmltopdf falls with NLP exceptions. You have may be a good common ideas, but Implementation leaves much to be desired.  

P.S. and think about such thing. I have specially added only 1 method with 1 line of code and changed another 1 line of code and no more. I added my own class, but I tried to leave the library and its code in its original form. No refactoring. No global changes. I have left even a monster constructor with 18 arguments, anyway my pull request will be most likely rejected... ) about extensibility / expandability - this library is almost not expandable.. and not because API, but IMHO: because of heavily use of structured programming patterns instead of classic OOP.

**e.G. I'd like to load resources not even from classpath, but from DB / database or through custom, not implemented in java protocol like SFTP.. With flying saurcer I can write a few lines of code to extend default UserAgent, but in openhtmltopdf you have only URL resolver.  Could you suggest any solution how to load resource via URL from DB ? ) or password protected resource, or something else ...**

rototor:
@delafer I am *not* the owner of the project, @danfickle is. Regarding the constructor, I've created #199. Yes, that constructor is code smell - but a private internal API anyway...

If you have some stuff which worked without problems in FlyingSaucer but crashes here you should open issues specific to those problems. The code sample I posted in #196 is for resolving absolute / relative URLs without protocol on the class path. Currently OpenHTMLToPDF does not resolve files in the class path itself. It could use Thread.getContextClassLoader() for this, should maybe even work in a container context.

I do not want to argue with you. This is an open source project were people work on this in their free time for FREE. If you don't like they way the project develops you don't need to use it. Or you can fork it and develop it the way you think it would be better.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxUserAgentFactory.java
rototor: This library can / will be used in whatever context the users want. Also it might be used with Tomcat (or Jetty, which also reuses threads). If we already know that using ThreadLocals will cause problems if the user is not careful we should avoid it - especially in the public API. The goal should always be to design an API in such a way that the user can not shoot himself into the foot. Not everyone has the time / is willing to read the documentation carefully. Being surprised by some "out of PermGen" errors in production is not so funny - especially if you don't have any clue where this could come from, as you usually have may different libraries in use.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfRendererBuilder.java
rototor: @delafer  Sorry, but where is the ```unregisterPdfBoxUserAgentFactory()``` call? Oh, you forgot it. So a class loader leak is incoming on WAR reload ... This is what I mean by bad API design, if it's easy to forget something you *will* forget to do it... and the problem will only manifest after the n-th reload of the WAR..

Yes, creating an idiot proof API is hard, but should be the goal never the less.

Method chaining means: ```builder.setXY().setAbc().setWhatever().build()```. For another example look here: https://google.github.io/gson/apidocs/com/google/gson/GsonBuilder.html

danfickle:
OK, firstly, let’s keep it nice and professional.

Secondly, thanks @delafer for your contribution. I broadly agree with most of your critique, I have to focus more on refactoring and testing than adding ever more features. One of the problems with a project of this size with very few tests, refactoring is fraught with danger. In hindsight I should have written more tests at the beginning.

In regard to the npe on uri resolving, it should work provided null isn’t passed in as the base uri. If it isn’t please post an issue with minimal example code.

In regard to this pull-request, while I try to accept most prs, I think that overriding the user agent is not a good idea. It is currently too messy and therefore unstable, so would break user code too often (when we changed the interface). For example, the caching system needs a big overhaul.

Now there are two ways we could enable your very valid use-case. We could refactor the pluggable Http/s stream resolver and make it for all protocols. Ie, the user passes in protocol(s) and a stream factory to use for that protocol. Alternatively, we could bring forward the cache work. As detailed in #170 the multi threaded cache would be backed by a concurrent hash map and store a uri to byte array mapping. It could be pre loaded with whatever the user required.

What does everyone think of these options?

Again, let’s keep it nice.

danfickle:
I’m going to close this pr, but feel free to keep civil discussion in this thread or on #196

rototor:
Regarding custom protocols I would suggest an API like this:

```
interface FSUrlOpener {
   InputStream openStream(URL url) throws IOException;
   Reader openReader(URL url) throws IOException;
}
```
with a default implementation like this:
```
public final class DefaultUrlOpener implements FSUrlOpener {
  ... _streamFactory;
   InputStream openStream(URL url) throws IOException {
       	if (url.getProtocol().equalsIgnoreCase("http") ||
		url.getProtocol().equalsIgnoreCase("https")) {
		return _streamFactory.getUrl(uri).getStream();
	}
       return url.openStream();
   }

   Reader openReader(URL url) throws IOException {
	if (url.getProtocol().equalsIgnoreCase("http") ||
		url.getProtocol().equalsIgnoreCase("https")) {
		return _streamFactory.getUrl(uri).getReader();
	}
       InputStream is = url.openStream();
       return is == null ? null : new InputStreamReader(is, "UTF-8");
   }
}
```

and an builder method to define custom UrlOpener:
```
builder.useUrlOpener(new FSUrlOpener(){
  // ... implement both methods and inspect url.getProtocol() to see what kind of transport you need.
  // You can always call url.openStream() if you don't care about a protocol and 
  // just want it to work (e.g. for class path resources)
});
```

If and how we cache the data returned here is an implementation detail, which we can change later. 

(Regarding DefaultUrlOpener being final: It requires the streamFactory, so a user deriving from DefaultUrlOpener would be forced to provide some streamFactory even if he does not care. I am not happy with that, better ideas are welcome)

So a user could customize how the URIs are built using ```useUriResolver()``` and customize how this URIs are opened using ```useUrlOpener()```.

danfickle:
@rototor Just saw this, but my implementation turned out similar! Feel free to suggest changes. Thanks.

rototor:
@danfickle I think that is fine. I would just suggest renaming HttpStream(...) to FSStream(...) as it's not HTTP only anymore.

delafer:
4ad3f0dc9b8db3eb0e3350d731dd7a24_400x400 1

PR #199, opened 2018-04-08
Builder State to transfer parameters

rototor:
The item, that the *Renderers take way to much arguments in #196 is valid. To solve this I extracted the builder state in its own class, which is then passed to the renderer.

This allows to get more state / parameters into the builder without extending the constructors all the time. That those state classes are public is a pity, but we don't have C++ friend classes at hand here...

This also adds the TestcaseRunner to the regression tests, just as a "smoke" test.


rototor:
Feel free to suggest a better class / variable name for the state.

danfickle:
Thanks a lot mate. Looks good.

delafer:
4ad3f0dc9b8db3eb0e3350d731dd7a24_400x400 1

PR #215, opened 2018-04-24
Fix font filehandle leak

rototor:
We've got a file handle leak in production which caused our Tomcat to stop working after some days, and were able to track this down to fonts files which were not closed. After the user generated enough reports the 2048 file descriptors were exhausted and everything stopped working ..

If a font file was registered but not used in the PDF, those font files would leak their file handles.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/SharedContext.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/SharedContext.java
@@ -34,7 +34,6 @@ import com.openhtmltopdf.util.XRLog;
 import org.w3c.dom.Document;
 import org.w3c.dom.Element;
 import org.w3c.dom.Node;
-import org.w3c.dom.NodeList;
 
 import java.awt.*;
 import java.text.BreakIterator;
@@ -165,10 +164,6 @@ public class SharedContext {
      */
     public FontResolver getFontResolver() {
         return fontResolver;
-    }
-
-    public void flushFonts() {
-        fontResolver.flushCache();
     }
 
     /**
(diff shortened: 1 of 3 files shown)


danfickle:
Thanks @rototor 

Are you saying only fonts loaded but not used leak? If so, perhaps this is a bug in PDFBOX. Anyway, is it possible to use the workaround mentioned in comments at link below, rather than calling subset? Ie. Get a ```TrueTypeFont``` object and close that in your close method?

https://stackoverflow.com/questions/35983135/pdfbox-leaves-font-files-open-after-generating-pdf

rototor:
@danfickle  Yes, that should theoretically work too. But it may throw an NPE as close() in TrueTypeFont is not allowed to be called more then once (which is a violation of the close() contract, but ...). To be exact RAFDataStream.close() will crash if it is called twice. Using subset() is a workaround for this problem... 

Maybe we should merge this ugly workaround and open a upstream ticket to fix close() in fontbox? And as soon as a fix is released we can cleanup this here.

danfickle:
OK, I'll file an issue with PDFBOX and merge this now. Thanks again.

THausherr:
@danfickle did you ever create an issue in PDFBox about this?

danfickle:
@THausherr 

No, I failed to get around to it. Should I file one now? By the way, huge thanks for your work on PDFBOX.

THausherr:
No longer needed, see https://issues.apache.org/jira/browse/PDFBOX-4242, please read it and comment if you want. I was just wondering whether we had missed an issue.

danfickle:
Thanks @THausherr 

I think that discussion covers it. In regard to this project, I was thinking we are only opening unused fonts for their font metrics, so we should probably cache this information between runs and only load the font if actually used in a document. This would be a performance improvement as well as a workaround for this issue.

Issue #217, opened 2018-04-26
NPE when stylesheet is not found


logsworth:
The DefaultHttpStreamFactory in NaiveUserAgent produces a DefaultHttpStream with null strm when it can't load the URI. The NPE is then thrown when you call getReader on the DefaultHttpStream.

PR #218, opened 2018-04-26
Update NaiveUserAgent.java

logsworth:
new InputStreamReader will throw a NPE if strm is null


logsworth:
for #217

danfickle:
Thanks a lot @logsworth 

Merging now.

Issue #80, opened 2017-04-03
Feature differences between openhtmltopdf and flyingsaucer


seizmo:
It would be greatly appreciated if you could add a section to the README that details the differences between openhtmltopdf and flyingsaucer in terms of features (apart from using libraries with better licenses).

 Are you behind, on par or even exceeding the features and HTML/CSS-compatibility of flyingsaucer?

rototor:
HTML/CSS compatibilty should be very similar. It seems flyingsaucer gets a little bit of development now again, the project was very quiet for a long time. 

As you may already know: Don't expect to throw any HTML on this library and get perfect printouts from it. To get good results you must tune your HTML to this library and use some CSS extensions. Many HTML5 features like "display:flex" wont get into this library nor flyingsaucer, because they are very complex to implement and you can already "fake" them using tables. Yes, tables are bad in normal HTML. But here you want to get a fixed width PDF printout, so you need to use whats there. On the other side their are extensions like ```-fs-table-paginate: paginate; ``` which allows you to get multipage tables with repeated headers. This is something which normal browser can't do. They usually put pagebreaks at random points which makes them ugly.

The features I added to openhtmltopdf are features which I needed:

- #31 / #32 Fixes/Workarounds für Pagebreaking. Only needed if you make complex multipage reports.
- A very little bit of transform:rotate(). But not really working 100% yet.
- #78 The possiblity to draw custom "objects", so you can draw directly into the PDF using any Graphics2D taking (chart/graph) library. 
- #23 Inline SVG Images

See the README for other changes. Currently the differences between flyingsaucer and openhtmltopdf are not that big. Beside using the old unmaintained iText2 vs active developed OpenPDFBox.

rajaningle:
In my opinion I can see one major difference between the two is Flying saucer with itext gives the ability to add pages to the pdf incrementally. That means it will allow you to create pdf and add one or more pages as you stream data and that seems memory efficient as well as fast.
openhtmltopdf is lacking on that front and don't support page addition which makes it render the entire thing at once which is causing performance issues. 
openhtmltopdf team is doing a great job in resolving this open issue. [180](https://github.com/danfickle/openhtmltopdf/issues/180)
but apart from this i think both libraries are similar and really good at generating pdf from HTML/CSS I couldn't find other library which supports the same and as feature rich as these two.
@rototor Flying saucer now supports itext5.

rototor:
[...]

Issue #214, opened 2018-04-22
Render PDF from multiple HTML files


sake92:
Hi all, is it possible to render a PDF from multiple input files/strings, like in [this example](https://github.com/flyingsaucerproject/flyingsaucer/blob/master/flying-saucer-examples/src/main/java/PDFRenderToMultiplePages.java) from flyingsaucer?  
I had a problem with flyingsaucer, it throws something like "Page 21 was requested but document has only 20 pages"... 😞  
I could set a `baseURL` for every page, on the `setDocumentFromString` method.  

With openhtmltopdf I should concatenate all the HTMLs into one, right? But the base folder is not the same for all of them... 😄  In some of them it is "../styles/main.css" but in some other "../../styles/main.css" (deeper folder)

danfickle:
This will work, sort of. The only problem is the ```counter(page)``` and ```counter(pages)``` will be wrong on subsequent documents. I'll try to get these working again in the fast renderer that I'm working on as part of #180.

````java
	//214
	public static void main(String...args) throws Exception {
		String[] uris = new String[] {
				"file:///Users/me/Documents/pdf-issues/issue-206.htm",
				"file:///Users/me/Documents/pdf-issues/issue-180-p.htm"
		};
		
		PDDocument doc = new PDDocument();
		
		for (String uri : uris) {
			PdfRendererBuilder builder = new PdfRendererBuilder();
			builder.withUri(uri);
			builder.usePDDocument(doc);
			PdfBoxRenderer renderer = builder.buildPdfRenderer();
			renderer.createPDFWithoutClosing();
			renderer.close();
		}

		OutputStream os = new FileOutputStream("/Users/me/Documents/pdf-issues/output/mytest-214.pdf");
		doc.save(os);
		os.close();
	}
````

sake92:
Thanks @danfickle, it works!!! 😄  

Although, I have some issues with characters like `č`,`ć` etc. They get turned into `#` character.

Also, code highlighting with http://prismjs.com/ isn't working same like in the browser.  
Seems like it "sees" just the `<code>` markup, not the prismjs goodies that get inserted after parsing the HTML.

danfickle:
Hi @sake92 

For the characters, you will still have to embed a valid font for most languages other than English. See the template author's guide on the readme for tips.

[...]

PR #222, opened 2018-05-22
#80 Fix and sample multihtmldoc pdf

rototor:
This is a sample testdriver how to use multiple HTML documents for one PDF. I also fixed a bug with bookmarks related to multiple HTML documents in one PDF.

The bookmark-markup is very special and esoteric (i.e. non-standard). But at least it works now again. No idea if we should let it be as it is, or come up with something more standard... Also requiring the bookmarks in the header is rather difficult if you only know the content at the end of the document...

Should PdfBoxRenderer::createPDFWithoutClosing() stay public, or should we try to hide it with the builder? i.e. as a builder option? This would also help the user discovering this feature.

### DIFF
--- a/openhtmltopdf-examples/src/main/java/com/openhtmltopdf/testcases/TestcaseRunner.java
+++ b/openhtmltopdf-examples/src/main/java/com/openhtmltopdf/testcases/TestcaseRunner.java
@@ -192,7 +192,7 @@ public class TestcaseRunner {
 		}
 	}
 
-	private static DefaultObjectDrawerFactory buildObjectDrawerFactory() {
+	static DefaultObjectDrawerFactory buildObjectDrawerFactory() {
 		DefaultObjectDrawerFactory objectDrawerFactory = new StandardObjectDrawerFactory();
 		objectDrawerFactory.registerDrawer("custom/binary-tree", new SampleObjectDrawerBinaryTree());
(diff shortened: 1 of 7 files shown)


danfickle:
Thanks @rototor 

There is another example in #214 
I suspect the page counter support with multiple documents has rotted however.

In regard to the bookmark mechanism, I did implement a standard based mechanism(based on a css property that would specify the bookmark level of an element) in neoflyingsaucer, so perhaps I could find that and incorporate it here.

```createPDFWithoutClosing``` should be deprecated in favour of a builder method as you suggest.

kingds:
Hi @rototor @danfickle 
Could you give some pointers on how to use bookmarks to link between multiple HTML files in the same PDF? 
I'm currently able to build a PDF using the method described by @danfickle here: #214
But I'm lost as to how I can create links between the different pages.

Thanks

danfickle:
Hi @kingds,

I don’t think it is possible. Generally it is better to combine the HTML in one file so fonts are subset only once, reused images embedded only once, etc. Could you outline your use case for why you need to do it in multiple runs?

kingds:
Okay thanks, I managed to get it working by combining all the HTML files into one document. 

The goal is to compile a JavaHelp helpset (bunch of HTML files) into a single PDF, based on the map file for the helpset.

PR #243, opened 2018-07-06
Fix #242 - Image with CSS max-width and max-height incorrectly scaled up

### LINKED ISSUE #242: Image with CSS max-width and max-height incorrectly scaled up
When the image supplied is smaller than both the max-width and max-height, the image is scaled up. Css max attributes should only result in down scale. When only one of the two max attributes is supplied, the up-scaling does not occur.

```
<html><head></head><body>
<img style="max-height: 950px; max-width: 710px" src="https://pbs.twimg.com/profile_images/638747380494503936/C89d1hl3_400x400.jpg" />
</body></html>

```

koan00:
Fix #242 - Image with CSS max-width and max-height incorrectly scaled up


danfickle:
Thanks a lot @koan00 

Doing a release now.

PR #250, opened 2018-07-13
Cherry-pick from PR #143: pdf/a support

syjer:
Hi @danfickle , I've cherry-picked from the PR done by @backslash47 the pdf/a support.

Additionally, from the PR, I've done some changes:
 - refactored the code: the 2 new parameters are set in the State object
 - switched from string to enum for setting the pdf/a level as requested in the comment of the PR
 - added the "StructTreeRoot" for passing the pdf/a-1 validation.

I've validated the various level of compliance using https://www.pdf-online.com/osa/validate.aspx and they seems to pass. Obviously you need to supply a ICC profile (I've used the one from https://svn.apache.org/viewvc/pdfbox/trunk/examples/src/main/resources/org/apache/pdfbox/resources/pdfa/ )

As an interesting note: for level 1/2 validation, you cannot use the fallback fonts: all the fonts must be embedded.

### DIFF
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxRenderer.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxRenderer.java
@@ -39,7 +39,6 @@ import com.openhtmltopdf.render.BlockBox;
 import com.openhtmltopdf.render.PageBox;
 import com.openhtmltopdf.render.RenderingContext;
 import com.openhtmltopdf.render.ViewportBox;
-import com.openhtmltopdf.render.AbstractOutputDevice.ClipInfo;
 import com.openhtmltopdf.render.displaylist.DisplayListCollector;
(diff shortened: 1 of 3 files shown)


danfickle:
Awesome. Thanks @syjer 

The only change I will make after the merge is to add NONE to the enum in place of using null to signify the same.

Also I will change the font resolver to not add the built-in fonts when PDF/A conformance is required as part of #249 

Thanks again.

syjer:
@danfickle thank you. 

By the way, if you choose the conformance level PDF/A-3, it's not required to embed the fonts, so the condition should be (NONE || PDF_A_3) :)

danfickle:
@syjer
I couldn’t find any info regarding a font embedding exception for pdf/a-3. Do you have a source? Thanks.

syjer:
@danfickle , I derived empirically using the pdf/a validator ( https://www.pdf-online.com/osa/validate.aspx ), but you are right that in the specifications, there is no exception. So most likely I'm wrong :)

Issue #237, opened 2018-06-26
Processing HTML with SVG image from remote URL throws IOException: Unrecognized Image format


mcbeelen:
I have a HTML-file, which contains an img with public URL to an SVG-image.

```
<p><img src="https://docs.chef.io/_images/start_chef.svg" alt="Chef Flow" /></p>
```

The PdfRendererBuilder is configured with the BatikSVGDrawer.

```
 builder.useSVGDrawer(BatikSVGDrawer())
```
When I try to create the PDF I get:
```
WARNING: Can't read image file; unexpected problem for URI 'https://docs.chef.io/_images/start_chef.svg'
java.io.IOException: Unrecognized Image format
	at com.openhtmltopdf.pdfboxout.PdfBoxImage.<init>(PdfBoxImage.java:53)
	at com.openhtmltopdf.pdfboxout.PdfBoxUserAgent.getImageResource(PdfBoxUserAgent.java:103)
	at com.openhtmltopdf.pdfboxout.PdfBoxReplacedElementFactory.createReplacedElement(PdfBoxReplacedElementFactory.java:59)
```

syjer:
Hi @mcbeelen 

Currently, you cannot link a svg from a img tag.

You need to embed the svg directly. And in this case,  you will need to add 

```
<dependency>
  <groupId>org.apache.xmlgraphics</groupId>
  <artifactId>batik-codec</artifactId>
  <version>1.9</version>
</dependency>
```

as the svg contains a base64 encoded image.

taras19921:
Hi all! My svg image is not displayed on PDF file. I have svg image on my html template. This is a snippet of my code:

`PdfRendererBuilder builder = new PdfRendererBuilder();
			builder.defaultTextDirection(BaseRendererBuilder.TextDirection.LTR);
			builder.useSVGDrawer(new BatikSVGDrawer());
			builder.useMathMLDrawer(new MathMLDrawer());
			builder.addDOMMutator(LaTeXDOMMutator.INSTANCE);
			builder.useDefaultPageSize(210, 297, BaseRendererBuilder.PageSizeUnits.MM);
			builder.withHtmlContent(html, "");
			builder.toStream(os);
			builder.run();` 

What I'm doing wrong or what I'm missing?

syjer:
hi @taras19921 , could you provide the html file?

The code seems ok.

taras19921:
Also I'm getting the following error: 

[...]

PR #254, opened 2018-07-14
implement #237 add support for svg files in img element

syjer:
Hi @danfickle , with this PR, we can use svg files in img tag.

I'm not sure if it's the best way, but it seems an unobtrusive change, so it feel kinda right :)

Additionally, I've added `batik-codec` as a dependency, so we can support base64 encoded images in the svg (as in the case of the svg linked in the issue #237).

I've tested with both the java2d and pdfbox backend. 

Note: the dimension of the rendered svg (both as a `svg` and `img` element) is not handled like chrome or FF. But currently I don't know why. You can try with `https://docs.chef.io/_images/start_chef.svg` .

Rendered by pdfbox:

pdfbox

Rendered by browser:

chrome

The code of the page is simply:

```
<html>

<body>
top
<p><img style="border:1px solid red;" src="https://docs.chef.io/_images/start_chef.svg" alt="Chef Flow" /></p>
bottom
</body>
</html>
```

edit: ok, there is already an issue (#128) I'll try to investigate then.

### DIFF
--- a/openhtmltopdf-java2d/src/main/java/com/openhtmltopdf/java2d/Java2DReplacedElementFactory.java
+++ b/openhtmltopdf-java2d/src/main/java/com/openhtmltopdf/java2d/Java2DReplacedElementFactory.java
@@ -36,10 +36,16 @@ public class Java2DReplacedElementFactory extends SwingReplacedElementFactory {
 			return new Java2DSVGReplacedElement(e, _svgImpl, cssWidth, cssHeight, box, context);
 		} else if (nodeName.equals("object") && _objectDrawerFactory != null) {
 			FSObjectDrawer drawer = _objectDrawerFactory.createDrawer(e);
-			if (drawer != null)
+			if (drawer != null) {
 				return new Java2DObjectDrawerReplacedElement(e, drawer, cssWidth, cssHeight,
 						context.getSharedContext().getDotsPerPixel());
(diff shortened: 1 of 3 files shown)


rototor:
Browsers usually use the content type and/or sniffer the file content to detect which "renderer" is needed for an image. That would be the "correct" way to handle this, but also would be way more complex to implement. So as a start I think it is ok to sniffer on the file extension.

As long as someone does not work with a content management system, which does not provide URLs with the right extensions, this should work fine.

Looks good to me.

syjer:
@rototor Yes, my initial idea was to check the content-type. but I didn't find any helpers/method in the `UserAgentCallback`.

danfickle:
I think that in the spirit of “perfect is the enemy of good enough” that using the url extension is ok. My only other concern is that allowing the use of external svg images might imply that arbitrary svg files are safe, which is not the case. However, we can address that in the documentation which I think I will move to a wiki to allow easier upkeep.

Anyway, huge thanks to @syjer and I’ll merge this now.

As to the size issue, I believe it defaults to 400 x 400 while the browsers default to a width of 100% with height determined by implicit aspect ratio. Possibly we could achieve this with a entry in the default style sheet. Can’t test now but will look later.

Also thanks to @rototor for reviewing.

PR #262, opened 2018-08-17
cli module for the project - easy adoption for non-tech (java) people 

ieugen:
Hello, 

I'm working with a front-end developer on producing PDF's documents using openhtmltopdf. 
Because he does not have technical skills I have created a simple CLI app to generate PDF from markdown files. 

Right now it's mostly a POC - can generate pdf from markdown and it can be expanded. 

This is very useful for non technical people to use this software. 

I would like to contribute this initial code to the project and if there is interest (and feedback) improve upon it over time. 

P.S. I am considering migrating this to gradle as it has much more powerful building capabilities for binary distribution.

Some future enhancements:

- add a server mode ? 
- render html

Feedback is welcome. 

```
java -jar target/openhtmltopdf-cli-0.0.1-RC16-SNAPSHOT-jar-with-dependencies.jar --help
Usage: render [-hV] [-i=FILE] [-o=FILE] [-r=FILE]
OpenHtmlToPdf - Render markdown document to PDF.
  -h, --help         Show this help message and exit.
  -i, --input=FILE   File to convert to PDF. Can be any supported file format:
                       Markdown. Defaults to README.md
  -o, --out=FILE     File to convert to PDF. Can be any supported file format:
                       Markdown.
  -r, --markdown-header=FILE
                     Html file to be used as header for markdown content.
  -V, --version      Print version information and exit.
```


ieugen:
There is also the option of publishing this as a package on https://sdkman.io/ . 
That will help distribution.

danfickle:
G’day @ieugen,

Welcome to the project and thanks for the contribution! Definitely, I think a cli is a good idea.

Unfortunately, some users still are not using Java 8+ so we need to keep testing on Java 7 jdk. Agree that rendering of HTML would be desirable. When you say server mode, do you mean like a json api?

We can keep working on this together. Next step is java 7 compatibility so it won’t break automated testing.

Thanks again,
Daniel.

ieugen:
Hi, 

I think the CLI should live a separate git project. That way we avoid the Java 7 compatibility issues (cli is consumer app, no need for that low compatibility). Also I'm thinking of using gradle - since building binary apps is much easier with gradle.  

Serve mode might not be such a good idea. I'm thinking of starting a web server - and watch files in a specific dir.

```
.
├── documentation
│   ├── documentation-header.htm
│   └── documentation.md
    └── html
        ├── overflow.html
        ├── positioning-absolute.html
        ├── positioning-fixed.html
```

Asking for: `http:localhost:8080/html/overflow.html` will render the HTML in browser.
Asking for something like `http:localhost:8080/html/overflow.html?print-pdf` witll render the PDF for that file. 

Similar thing for the markdown stuff. 

WDYT?

danfickle:
Yes, a separate repository might be the way to go. Are you happy to host it? A local server seems like a cool idea.

ieugen:
I can host the project and publish it on sdkman + instructions on how to install and use it.

PR #284, opened 2018-10-15
Update batik deps due to CVE-2018-8013

ghenadiibatalski:
Our security engineer stumbled over the [CVE-2018-8013](
https://nvd.nist.gov/view/vuln/search-results?adv_search=true&cves=on&cpe_version=cpe%3A%2Fa%3Aapache%3Abatik%3A1.9)

I increased the required version to 1.10 and run into the issue [BATIK-1229](https://jira.apache.org/jira/browse/BATIK-1229) so the dependency to batik-ext seems to be necessary to work around this issue.

### DIFF
--- a/openhtmltopdf-svg-support/pom.xml
+++ b/openhtmltopdf-svg-support/pom.xml
@@ -39,12 +39,12 @@
     <dependency>
 	    <groupId>org.apache.xmlgraphics</groupId>
 	    <artifactId>batik-transcoder</artifactId>
-	    <version>1.9</version>
+	    <version>1.10</version>
     </dependency>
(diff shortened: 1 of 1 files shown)


danfickle:
Thanks a lot. It should be noted that untrusted svg images should not be used due to the ability of svg to include file uris to incorporate raster images. At the very least, this gives it the ability to probe the file system. Anyway, thanks again, merging now.

danfickle:
P.S I’ll do a release over the weekend.

danfickle:
RC17 just released. Thanks @ghenadiibatalski.

Issue #280, opened 2018-10-08
Move to Java 8 requirement?


danfickle:
We were thinking of moving to a minimum of java 8. This would give us streams/lambdas and other modern goodies. It’s been several years since java 8 was released. If you object, here is your chance to say why.

achuinard:
No objection. Great work with this open source library, @danfickle. Sustain its future with a move to Java 8.

Fancellu2:
Agreed. Java8 has been out for 4 years.

If someone is stuck at jdk 7, they can always use an older release

dilworks:
No problem here - been on Java 8 for a while without incidents. in fact now that J2EE 8 is a thing, I'm looking forward to move to Java 9 (or whatever is the latest version for this week - this new versioning scheme is really awful, thanks Google!)

If you're stuck with Java 7, hope you're paying those megabuckos to Oracle for that support contract!

danfickle:
RC18 is Java 8 all the way, so closing this.

PR #300, opened 2018-12-01
Update project to compatiple with Java 6

dnguyenminh:
Hi Danflickle,

I have update the project to make it compatible with Java 6.


danfickle:
Hi @dnguyenminh,

I’d like to thank you for this work and your interest in the project.

Unfortunately, your pull request arrived only a day after we deliberately made the move to Java 8. The move to 8 was made after a lengthy consultation period #280 , that no one objected to.

Java 8 has been out for several years and I still think now is the right time to make the upgrade.

Therefore, I can not accept this request. Again, thanks for your interest and please don’t let this put you off making future contributions, although you can always propose in an issue first.

Regards,
Daniel.

PR #301, opened 2018-12-04
Make loading resources from classpath work when openhtmltopdf is a named module

zimmi:
Using `Class::getResource*` works on both module- and classpath (see [here](
https://stackoverflow.com/a/45173837)), the code was using `ClassLoader::getResource*` previously.

I added some tests to verify the classpath behavior still works. Still not sure how to test the module path behavior properly though.


danfickle:
Thanks @zimmi. Looks good.

PR #304, opened 2018-12-07
Fix warnings for icon font without space inside PDF/A, add tests

zimmi:
Googles MaterialIcons-Font does not contain a space character. When using PDF/A, the layout algorithm tries to eagerly find out the width of the space character. This does not happen when PDF/A is not used.
I changed the algorithm to only query for the width if there is actually a space inside the element.

Added tests to verify the fix (the PDF/A one failed previously).


danfickle:
It failed because we always use the built in serif as a fallback if PDF/A conformance is not required. Your code looks fine so merging now. Thanks!

Issue #319, opened 2019-01-26
PdfRendererBuilder - Incorrect font configuration throws NPE


brundipub:
Current status: When a font is incorrectly configured the app just throws a NullPointerException in PdfBoxTextRenderer.

There might be various reasons of this failure however the exception is still the same:
https://github.com/danfickle/openhtmltopdf/issues/290
https://github.com/danfickle/openhtmltopdf/issues/277

Expected: There should be a warning about the font issue in log and the app should continue in rendering using another (default) font.

danfickle:
Thanks @brundipub,

Looks good. Better failure modes are always a good idea. Could you possibly add a commit that also warns if the font list is empty? This can happen for PDF/A or PDF/UA documents where the built-in fonts can not be used.

Thanks again.

brundipub:
I cannot test it on a real data but basic check has been added to prevent empty list.
Anyway the previous modif has been tested with incorrect font file path and it works as expected.

PR #320, opened 2019-01-26
#319 - NPE prevention in case of incorrect font configuration

brundipub:


danfickle:
Awesome, thanks @brundipub.

PR #333, opened 2019-03-04
Upgrade PDFBox and PDFBox-Graphics2D versions

rototor:
@danfickle I think it would be a good idea to upgrade the PDFBox version before releasing a new version.


rototor:
Oh, my new pdfbox-graphics2d version is not on maven central yet ... I'll push a readme update for this changes into this branch in an hour, then the build should pass.

danfickle:
Thanks @rototor. You’re always one step ahead of me.

PR #334, opened 2019-03-11
Control of PDPage creation via PDFCreationListener

DSW-PS:
Hello,

I think it would be a good idea to give the user more control over how a `PDPage` is created and added to the `PDDocument`.

I modified the `PdfBoxRenderer` to delegate the page-creation to the `PDFCreationListener` (if one is set).

(In our case we already have `PDPage` with some data created and now need to print the rest - we could of course just print the html to pdf to import and place the document, but this way it is only a matter of setting where the html is to begin via some styling rules, much more simpler and streamlined)

Kind regards
Peter


danfickle:
Hi @DSW-PS,

Thanks a lot for the PR.

I think that allowing the user to control page creation is a good idea. However, I’m not sure that using `PDFCreationListener` is the way to go. Firstly, I’m not sure that its methods are all being called in the right place, so that would need analyzing before adding it to the official documented api. Secondly, the modern trend is towards lambda compatible interfaces for callbacks.

So I would prefer a specific lambda compatible interface for the callback and the ability to set it from the builder. Maybe `usePageSupplier`. Finally we need the page number and shadow page number as part of the arguments.

What do you think? If you think this is reasonable, I could do it when time permits, or it would be great if this pr or another implemented it!

Thanks again,
Daniel.

P.S. Anyone have any idea why Travis ci hasn’t picked this pr up?

ieugen:
> P.S. Anyone have any idea why Travis ci hasn’t picked this pr up? 
Some events may be lost ?

DSW-PS:
Hi @danfickle,

I tried to understand what the shadow-pages are for (or how they are formed) but I can't say I understand.

Nevertheless I took your advice and removed the changes to the `PDFCreationListener` and instead put the (possible) creation of the `PDPage` in the new interface `PageSupplier` - which can be written as a lambda. 
Also added the `pageNumer` and `shadowPageNumber` as arguments as you suggested.

Tell me what you think,
Peter

danfickle:
Thanks @DSW-PS,

Shadow pages are there for [cut-off pages](
https://github.com/danfickle/openhtmltopdf/wiki/Cut-off-page-support). The pr is looking good. Only minor concerns:

Page number should be zero based (use `i`) and continue to use `i` In the shadow page loop for the page number argument. Other than that, really good.

DSW-PS:
Hi @danfickle,

thanks for the explanation. I added a brief comment with link to the [cut-off pages wiki page](https://github.com/danfickle/openhtmltopdf/wiki/Cut-off-page-support) to the `PageSupplier` to help others.

Now using `i` for `pageNumber` in the shadow-page loop.

danfickle:
Sensational. Thanks a lot @DSW-PS.

PR #373, opened 2019-07-29
Fix for a regression with image sizing when specifiying both a max-width

rototor:
and a max-height. This just did not work correctly, it broke the layout.

We must respect the scaling factor between instrinct-sizes and css-sizes.

This was already fixed in the "too small case", but not in the other case.

Feel free to replace the sample images with something else. Just make sure that they are big enough.


danfickle:
Thanks @rototor,

I don't know how I missed this. Even the comment above hints at the correct code!

Are the images OK to relicense as LGPL? If not I probably should change them.

rototor:
@danfickle This are customer pictures, so it would be better if you could replace them.

Issue #383, opened 2019-09-06
Allow users to provide FSSupplier<PDFont> in PdfBoxFontResolver to avoid duplicate fonts/subsets


DSW-PS:
Hi @danfickle 

what do you think about adding a method to the **PdfBoxFontResolver** that allows users to use their own **FSSupplier<PDFont>**?

This way it would be possible to avoid having duplicate fonts (or fonts with several subsets) in the PDF by using some sort of font-cache.

This quite easily happens if you have _openhtmltopdf_ produce the output in a **PDDocument** where you also manually add some text which uses (some of) the same fonts as the html.
In this case it would be up to the implementation of **FSSupplier** to either load the **PDFont** or serve a already loaded one.

I can provide a pull-request for this.

danfickle:
Hi @DSW-PS,

I wouldn't be opposed to a pr. However, i think i'd prefer a method on the PDF builder if that's possible. In the meantime, you may be able to get added fonts from the page resource dictionary of the pddocument.

DSW-PS:
Hi @danfickle,
I created the pr.

I wasn't too keen on reordering the method parameters to keep the **FSSupplier<PDFont>** directly in the method as it would clash with the ones that use **FSSupplier<InputStream>**.
Instead I moved the **PDFontSupplier** from **PdfBoxFontResolver** to a seperate class that can be extended.

PR #384, opened 2019-09-09
useFont/addFont with PDFontSupplier (FSSupplier<PDFont>)

DSW-PS:
Hi @danfickle 
this is the pr I offered in issue #383 .

As per your suggestion I also added the functionality as _useFont_ methods to the **PdfRendererBuilder**.

I moved the already existing **PDFontSupplier** to a normal class so we can extend it - I would have prefered a _useFont_ / _addFont_ with direct **FSSupplier<PDFont>** but I wasn't too keen on reordering parameters just to clear the _type-earasure_ error as it would clash with the _useFont/addFont(FSSupplier<InputStream> ...)_ methods.


danfickle:
Thanks a lot @DSW-PS. All looks good.

PR #394, opened 2019-10-11
Update PDFBox to 2.0.17

cristan:
This makes a crash in case of an invalid font a lot nicer. Instead of a NullPointerException, you'll get a more descriptive IOException "Could not get hhea table".

Old stacktrace:
```
com.openhtmltopdf.load INFO:: Loading font(noto) from InputStream supplier now.
java.lang.NullPointerException
	at org.apache.fontbox.ttf.HorizontalMetricsTable.read(HorizontalMetricsTable.java:53)
	at org.apache.fontbox.ttf.TrueTypeFont.readTable(TrueTypeFont.java:353)
	at org.apache.fontbox.ttf.TTFParser.parseTables(TTFParser.java:173)
	at org.apache.fontbox.ttf.TTFParser.parse(TTFParser.java:150)
	at org.apache.fontbox.ttf.TTFParser.parse(TTFParser.java:106)
	at org.apache.pdfbox.pdmodel.font.PDType0Font.load(PDType0Font.java:94)
	at com.openhtmltopdf.pdfboxout.PdfBoxFontResolver$FontDescription.realizeFont(PdfBoxFontResolver.java:763)
	at com.openhtmltopdf.pdfboxout.PdfBoxFontResolver$FontDescription.getFontMetrics(PdfBoxFontResolver.java:828)
	at com.openhtmltopdf.pdfboxout.PdfBoxTextRenderer.getFSFontMetrics(PdfBoxTextRenderer.java:76)
	at com.openhtmltopdf.layout.LayoutContext.getFSFontMetrics(LayoutContext.java:425)
	at com.openhtmltopdf.css.style.CalculatedStyle.getFSFontMetrics(CalculatedStyle.java:771)
	at com.openhtmltopdf.css.style.CalculatedStyle.getLineHeight(CalculatedStyle.java:464)
	at com.openhtmltopdf.layout.InlineBoxing.layoutContent(InlineBoxing.java:96)
	at com.openhtmltopdf.render.BlockBox.layoutInlineChildren(BlockBox.java:1194)
	at com.openhtmltopdf.render.BlockBox.layoutChildren(BlockBox.java:1175)
	at com.openhtmltopdf.render.BlockBox.layout(BlockBox.java:1032)
	at com.openhtmltopdf.render.BlockBox.layout(BlockBox.java:947)
	at com.openhtmltopdf.layout.BlockBoxing.layoutBlockChild0(BlockBoxing.java:321)
[...]

PR #395, opened 2019-10-12
Update pdfbox-graphics2d version to 0.25

rototor:
And while we are at it, we should also update PDFBox itself. But this is redundant with #394.


danfickle:
Thanks @rototor

PR #400, opened 2019-10-29
lang and title support for span tag

Ignaciort91:
We are working on PDF's with accessibility support and we find two use cases not contemplated until the moment in the span tags:

- Alternative text
In some cases it is necessary to describe an acronym as Kms: Kilometers.

- Change language with leng
https://accessibility.psu.edu/foreignlanguages/langtaghtml/
NOTE: The change was made only in the span tag, but a generic could be implemented for all tags.

### DIFF
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxAccessibilityHelper.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxAccessibilityHelper.java
@@ -75,7 +75,8 @@ public class PdfBoxAccessibilityHelper {
         suppliers.put("th", TableHeaderStructualElement::new);
         
         suppliers.put("a", AnchorStuctualElement::new);
-        
+        suppliers.put("span", SpanStuctualElement::new);
+
         return suppliers;
     }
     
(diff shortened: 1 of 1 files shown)


danfickle:
Thanks a lot for putting this effort in @Ignaciort91. Much appreciated.

After some research it seems that the PDF reference recommends using the expanded entry for acronyms and abbreviations. The alternate entry is typically used for content with no textual equivalent such as images or symbol fonts.

The `abbr` tag is typically used for this purpose in HTML. As you point out the `title` and `lang` attributes can be used on any tag in html.

So here is what I'll do:
+ Special case the `abbr` tag (by copying, pasting and editing your code) to specify acronyms and abbreviations, setting the expanded entry on the structural element from the `title` attribute.
+ Make sure all tags handle the `title` and `lang` attributes. All tags, except `abbr` will use the `title` attribute to set the alternate text. Therefore, users will have the choice of using `abbr`, `span`, or another tag as appropriate.
+ Add `abbr` tag and `span` with `lang` and `title` attributes to the all-in-one PDF-UA test case.

Thanks again,
Daniel.

Ignaciort91:
Excellent! Thank you @danfickle for your prompt implementation and in a more generic way. I think that these types of improvements are very useful to help people with difficulties and make the library much more robust.

Issue #408, opened 2019-11-15
Bookmarks are not working with JSoup


Milchreis:
If you generate a PDF with bookmarks with JSoup the bookmarks are not included and no error message is thrown.

## Example
```java
String html = "<html>\n" +
        "<head>\n" +
        "<style>\n" +
        "div {\n" +
        "\tpage-break-after: always;\n" +
        "}\n" +
        "#toc {\n" +
        "    width: 100%;\n" +
        "    border-collapse: collapse;\n" +
        "}\n" +
        "#toc .page-number::after {\n" +
        "  /* SPECIAL STUFF HERE! */\n" +
        "  content: target-counter(attr(href), page);\n" +
        "  width: 30px;\n" +
        "}\n" +
        "</style>\n" +
        "<bookmarks>\n" +
        "  <bookmark name=\"Title of element on page 1\" href=\"#page-1\"/>\n" +
        "  <bookmark name=\"Title of element on page 2\" href=\"#page-2\"/>\n" +
        "  <bookmark name=\"Title of element on page 3\" href=\"#page-3\"/>\n" +
        "  <bookmark name=\"Title of element on page 4\" href=\"#page-4\"/>\n" +
        "</bookmarks>\n" +
        "</head>\n" +
        "<body>\n" +
        "<h1>Bookmarks and TOC example</h1>\n" +
        "\n" +
        "<h2>TOC</h2>\n" +
        "<table id=\"toc\">\n" +
        "  <tr><td><a href=\"#page-1\">Title of element on page</a></td><td class=\"page-number\" href=\"#page-1\"></td></tr>\n" +
        "  <tr><td><a href=\"#page-2\">Title of element on page</a></td><td class=\"page-number\" href=\"#page-2\"></td></tr>\n" +
        "  <tr><td><a href=\"#page-3\">Title of element on page</a></td><td class=\"page-number\" href=\"#page-3\"></td></tr>\n" +
        "  <tr><td><a href=\"#page-4\">Title of element on page</a></td><td class=\"page-number\" href=\"#page-4\"></td></tr>\n" +
        "</table>\n" +
        "\n" +
        "<div id=\"page-1\">Page 1</div>\n" +
        "<div id=\"page-2\">Page 2</div>\n" +
        "<div id=\"page-3\">Page 3</div>\n" +
        "<div id=\"page-4\">Page 4</div>\n" +
        "\n" +
        "</body>\n" +
        "</html>";

[...]

PR #409, opened 2019-11-15
handle bookmarks in body #408

syjer:
hi @danfickle , this PR should fix the issue #408 :).

It will search the "bookmarks" element both in the head and body.


danfickle:
Awesome. I very much like a PR with a test! Thanks a lot again @syjer.

Issue #366, opened 2019-07-08
Bolded numeration for no reason


acieplinski:
Hello,
I found some weird behaviour when I generate PDF with 2 or more pages. When there's a bold word (in < strong> tags, as you can see '< strong>expedita< /strong>' in my attachment) in the last line of the page, then numeration on the next page is bold as well (it shouldn't be as there's a < /strong>)
This is my code:
```
public void htmlToPdf(OutputStream outputStream, String htmlString) {
		try {
			PdfRendererBuilder builder = new PdfRendererBuilder();
			String xhtml = convertToXhtml(htmlString);
			builder.withHtmlContent(xhtml, "");
			builder.toStream(outputStream);
			builder.run();
		} catch (Exception e) {
			e.printStackTrace();
		}
	}
```

There's a 'convertToXhtml' method, I attached  xhtml.txt so you can review what is put into 'withHtmlContent(...)'
Have you met that problem? Any idea if I can avoid rendering pdf that way?

[xhtml.txt](https://github.com/danfickle/openhtmltopdf/files/3368090/xhtml.txt)
![image](https://user-images.githubusercontent.com/34013693/60806231-c0901800-a182-11e9-88d2-958d379b05d7.png)

syjer:
hi @acieplinski , you are right, it's a bug. 

I've tested in the sandbox (https://sandbox.openhtmltopdf.com/) and I can reproduce your issue.

As a simple (and temporary) workaround, you can do the following, add a custom style to override the counter:

```html
<style>
ol {
  list-style: none;
  counter-reset: list-counter;
}
ol li {
  counter-increment: list-counter;
}
ol li::before {
  content: counter(list-counter) ". ";
  font-weight: normal;
}
</style>
```

acieplinski:
Could you describe how do you 'inject' the styling? It should be in xhtml in header? Or there's a way to use an external stylesheet?

syjer:
@acieplinski  I think that I would add it inline in the document.

 When you get the html string, you convert it in a dom tree, create the style element, add it in the "head" element and you are good to go.


Or yes, you can reference an external css if you want.

syjer:
Taking a better look at this issue.

A shorter and simpler reproducer:

```html
<html>
<head>
    <style>
        p {
            page-break-after: always;
        }
    </style>
</head>
<body>
<p>
    <b>bolded text</b>
</p>
<ol>
    <li>counter should not be bold</li>
</ol>
</body>
</html>
[...]

PR #410, opened 2019-11-16
ListItemPainter: set font specification when calling drawText #366

syjer:
Hi @danfickle , this PR fix the issue described in #366 . 

I've added a visual regression test too (and checked that the test fail when disabling the fix :)).


danfickle:
Hi @syjer,

It looks like you have diagnosed the problem perfectly. Thanks for the fix (and the test)!

Issue #151, opened 2017-11-29
NPE if the name attribute for form element is not set


mmatecki:
Hello,
There's NPE if HTML form element doesn't have `name` attribute set.

Example HTML:
```
<!doctype html>
<html lang="en">
  <head>
    <meta charset="utf-8">    
  </head>
  <body>
  <form>
    <input type="text"/>
  </form>
  </body>
</html>
```
Exception:
```
java.lang.NullPointerException
	at com.openhtmltopdf.pdfboxout.PdfBoxForm$Field.access$602(PdfBoxForm.java:78)
	at com.openhtmltopdf.pdfboxout.PdfBoxForm.processTextControl(PdfBoxForm.java:435)
	at com.openhtmltopdf.pdfboxout.PdfBoxForm.process(PdfBoxForm.java:803)
	at com.openhtmltopdf.pdfboxout.PdfBoxOutputDevice.processControls(PdfBoxOutputDevice.java:379)
	at com.openhtmltopdf.pdfboxout.PdfBoxOutputDevice.finish(PdfBoxOutputDevice.java:1001)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.writePDF(PdfBoxRenderer.java:607)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.createPDF(PdfBoxRenderer.java:546)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.createPDF(PdfBoxRenderer.java:485)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.createPDF(PdfBoxRenderer.java:481)
	at com.openhtmltopdf.pdfboxout.PdfRendererBuilder.run(PdfRendererBuilder.java:86)
```

Regards,
Maciek.

iristic:
two years too late... but your input element needs to have name attribute set in order not to get NPE.

PR #413, opened 2019-11-17
handle more gracefully an input without name attribute + select without option child #151

syjer:
hi @danfickle , this PR handle more gracefully input/select/textarea elements without the name attribute.

Currently they launch a NullPointerException. With this PR, only a message at WARN level will be logged. 

Additionally, I've added a more lenient handling for select elements without option child. (same NPE issue).

This fix issue #151


danfickle:
A sincere thanks to @syjer for all these fixes to long standing issues!

PS. I'm about to do a release.

Issue #261, opened 2018-08-08
IndexOutOfBoundsException on empty div referenced as an anchor


ghost:
Hello-

Sample HTML file below will produce an index out of bounds exception.  I studied the code in depth and have figured out how to work around it but haven't quite figured out the cause.

**Stacktrace:**
Exception in thread "main" java.lang.IndexOutOfBoundsException: Index out of bounds: 2
	at org.apache.pdfbox.pdmodel.PDPageTree.get(PDPageTree.java:286)
	at org.apache.pdfbox.pdmodel.PDPageTree.get(PDPageTree.java:211)
	at org.apache.pdfbox.pdmodel.PDDocument.getPage(PDDocument.java:1392)
	at com.openhtmltopdf.pdfboxout.PdfBoxLinkManager.createDestination(PdfBoxLinkManager.java:356)

[PdfBoxLinkManager.txt](https://github.com/danfickle/openhtmltopdf/files/2272090/PdfBoxLinkManager.txt)
[sample.txt](https://github.com/danfickle/openhtmltopdf/files/2272091/sample.txt)

syjer:
hi @dmartineau99 , it's because there is no second page added if the div is empty, but internally it seems to have decided that the div is present and it's in the second page. (You can check that there is no second page by simply removing the anchor link in the Table of Contents).

So when he try to set the links, it pass the index of the non existing page and it will fail. A nice corner case.

edit: I wonder if the correct render of this page would be a second empty page...

ghost:
Probably so-- an empty page would be my vote.  I ran another test with the empty div followed by another div with content.  So theoretically it should have created 3 pages, one of which would be empty (middle).  It created 2 and did not fail.  So this particular issue is a trailing (end of the DOM) empty element.

On a bright note-- this project really rocks.  This particular HTML file is actually much larger than the stripped down one I put here.  I started with FS and it actually hangs (indefinitely) trying to process it.  This project processed it in just a few seconds.  I'll be doing a bit more testing then switch over.  My company is certainly willing to help contribute code-- its just we're Subversion people so it will take a bit to get spun up on Github and be useful.

PR #414, opened 2019-11-18
add test for issue #261. Confirm fast renderer does not launch exception

syjer:
Hi @danfickle , I've added a test for the issue #261 , the fast renderer does not launch the `IndexOutOfBoundsException` exception. (The old one still has the issue :))

The open issue is: given the html, should it create 1 or 2 pages ? Currently, it only create 1 page (coherent on both slow and fast renderer).


danfickle:
Thanks @syjer, I think boxes with ids should always generate a page, just in case they are used as a bookmark or link target.

Issue #309, opened 2018-12-13
ClassCastException AnonymousBlockBox cannot be cast to TableCellBox on TD: float-left


ghost:
We see this exception:
java.lang.ClassCastException: com.openhtmltopdf.render.AnonymousBlockBox cannot be cast to com.openhtmltopdf.newtable.TableCellBox

With HTML
`            <table>
                <tr>
                    <td style="float:left">something</td>
                    <td>xx</td>
                </tr>
            </table>
`
Granted float:left on a td doesn't make sense but a warning to the log is preferable to an exception if possible.

sandre1:
The same thing happens when using Html : 
`<td><img src="/img/regular/address-card.svg" width="20" /></td>`

danfickle:
Thanks again @syjer for fix. Closing for now.

PR #415, opened 2019-11-18
fix issue #309, avoid adding AnonymousBlockBox on TableCells. Additionally build a TableCellBox instead of a BlockBox if it's a floating element

syjer:
This PR fix the issue #309 . 

I'm not exactly sure if the change in `insertAnonymousBlocks` is the best way to prevent adding an AnonymousBlock, but it's the least intrusive one :)


danfickle:
Thanks @syjer. Merging now.

Issue #399, opened 2019-10-28
-fs-table-paginate tends to leave only thead and tfoot before page break


VictorAtPL:
I need to convert the html with multiple tables, which has thead, tbody, tfoot. The tables should be paginated if needed.

Here is the result I could achieve:
<img width="100" alt="openhtmltopdf_bug" src="https://user-images.githubusercontent.com/1886347/67725379-77d6c600-f9e2-11e9-9364-0db9531cbb3e.png">

I've already tried #354 with no success.

Could you please guide me what should I try and/or where can I find the code responsible for paginated tables? I will try to do some dirt fix for myself.

danfickle:
Yes, nested paginated tables are definitely buggy. I think it is related to the `ContentLimitContainer` (in the render package) and its use, but I don't fully understand the table code.

Another issue, which you may be able to solve is we don't have a minimal example showing the problem. An absolutely minimal sample would help a lot.

VictorAtPL:
> Yes, nested paginated tables are definitely buggy. I think it is related to the `ContentLimitContainer` (in the render package) and its use, but I don't fully understand the table code.

Do you know how this library works in overall? There is lack of information of it. I am debugging the code for the second consecutive day and still not sure how it works.

It looks like: HTML -> BlockBox -> Layer -> Ops -> PdfBox renderer? I still can't find the place in code where table is split across to pages. For me it looks like the table is queued to render on both pages, but there are some overlays/constraints which prevents the insufficient part of table to be rendered on proper page.

If I understand it right, there is one TableBox for each Table in html and later if there is need to split the page on multiple pages then more than one layer is created. Am I right?

> Another issue, which you may be able to solve is we don't have a minimal example showing the problem. An absolutely minimal sample would help a lot.

If I don't succeed to do it using my private code, I will try to prepare a _MWE_.

VictorAtPL:
Here is MWE:
https://gist.github.com/VictorAtPL/82e035e5f4e306aa692cb206d6624347

If you think I should make it even more minimal, I will try.
`background-color` and `border` statements can be of course removed if it will make debugging easier.

Preview:
![image](https://user-images.githubusercontent.com/1886347/69258174-7dc45f00-0bbc-11ea-8a94-73d5f1992f1f.png)

VictorAtPL:
I made a dirty fix in #416, although still hiding the border should be added and some test cases of other users should be performed.

[...]

Issue #202, opened 2018-04-10
Paginated tables that start at top of page have incorrect header layout


jesselong:
```
<html>
<head>
<style>
table {
    -fs-table-paginate: paginate;
//        page-break-inside: avoid;
}
th, td {
    border: 1px solid black;
}
</style>
</head>
<body>
<h1>Table test</h1>

<div>Hello1</div>
<div>Hello2</div>
<div>Hello3</div>
<div>Hello5</div>
<div>Hello6</div>
<div>Hello7</div>
<div>Hello8</div>
<div>Hello9</div>
<div>Hello10</div>
<div>Hello11</div>
<div>Hello12</div>
<div>Hello13</div>
<div>Hello14</div>
<div>Hello15</div>
<div>Hello16</div>
<div>Hello17</div>
<div>Hello18</div>
<div>Hello19</div>
<div>Hello20</div>
<div>Hello21</div>
<div>Hello22</div>
<div>Hello23</div>
<div>Hello24</div>
<div>Hello25</div>
<div>Hello26</div>
<div>Hello27</div>
<div>Hello28</div>
<div>Hello29</div>
<div>Hello30</div>
<div>Hello31</div>
<div>Hello32</div>
<div>Hello33</div>
<div>Hello34</div>
<div>Hello35</div>
<div>Hello36</div>
<div>Hello37</div>
<div>Hello38</div>
<div>Hello39</div>
<div>Hello40</div>
<div>Hello41</div>
<div>Hello42</div>
<div>Hello43</div>
<div>Hello44</div>
<div>Hello45</div>
<div>Hello46</div>
<div>Hello47</div>
<div>Hello48</div>
<div>Hello49</div>
<div>Hello50</div>
<div>Hello51</div>
<div>Hello52</div>
<div>Hello53</div>
<div>Hello54</div>

<table>
    <thead>
        <tr><th>HA</th><th>HB</th></tr>
    </thead>
    <tbody>
        <tr><td>a</td><td>b</td></tr>
        <tr><td>a</td><td>b</td></tr>
    </tbody>
</table>

</body>
</html>
```

At least on my machine, that puts the `table` on a new page, but the first `td`'s have too much space above them, like padding in the `td`. If you add `page-break-inside: avoid` to the `table` style in the `head` section, then the error is slightly different.

I think the problem has something to do with rendering both the original `th` and the paginate one.

PR #418, opened 2019-11-23
#399 #416 Initial fix for header/footer showing on page without body rows

danfickle:
I also plan to add some commits to this PR with general table cleanup of code.

@VictorAtPL, if you get time, you could perhaps test this branch with your real world tables.


danfickle:
The test I just committed for #202 still produces this:

issue-202-malformed-table

so clearly there is still more related problems in the table code.

VictorAtPL:
Right now it works differently. It puts part of text on the previous page, and part of it on the next, even that I have:
```
div.parts table.part tr, div.parts table.part tr td {
    page-break-inside: avoid;
}
```
in the styles.
openhtmltopdf_result_after_authors_bugfix

leedsalex:
@danfickle Hi, any update when this will be ready to be released? I'm hopeful it will fix the issues i'm experiencing.

danfickle:
@VictorAtPL, it is not clear from your comment if this is a problem introduced in this pr or was already existing? If it was already existing I propose to merge this pr and do a release, then we can work more on table rendering in a different pr.

@leedsalex Very soon if this pr has not introduced any additional problems as above.

VictorAtPL:
@danfickle 
I will test it tomorrow and answer your question.

VictorAtPL:
@danfickle
I am not sure. 

v. 1.0.1:
image
image

fix_399_empty_table_header_footer
image
image

https://github.com/VictorAtPL/openhtmltopdf/tree/fix_thead_tfoot_without_content_bug (my dirty fix)
image
image

As you can see, there are still problems with whole header and footer being not rendered ("rectangle" near "Grafika" and arrow below "C"). Also, on your branch, "dsadsa" of second C is on the previous page, and image of the second C doesn't have margin.

danfickle:
Hi @VictorAtPL,
I'm back finally. I had a play around but wasn't able to replicate. Are you able to share how the left blue arrow bar is achieved? At a guess it involves rowspan, transforms and a CSS arrow? Even better would be a sample.

Issue #420, opened 2019-11-27
text-align: justify causes NullPointerException when using <br /> in <span>-tags


dudeltrieb:
Hi,

I have a problem when using <br /> tags inside of span tags when the paragraph has text-align: justified. I'am using Verions 1.0.1 (no problem in 1.0.0). 

**HTML Code** (no other code is used is used to reproduce the problem)
```
<html>
  <head>
  </head>
  <body>
    <p style="text-align: justify">
      <span>foo<br />bar</span>
    </p>
  </body>
</html>
```

**StackTrace**
```
Caused by: com.myCompany.mySystem.documentrenderer.renderer.DocumentRendererException: Error when rendering pdf document.
	at com.myCompany.mySystem.documentrenderer.renderer.impl.DefaultDocumentRenderer.renderPdfDocumentByHtml(DefaultDocumentRenderer.java:83)
	at com.myCompany.mySystem.documentrenderer.renderer.impl.DefaultDocumentRenderer.renderPdfDocumentByHtml(DefaultDocumentRenderer.java:54)
	at com.myCompany.mySystem.documentrenderer.tags.RenderPdfTag.doEndTag(RenderPdfTag.java:90)
	at org.apache.jsp.tag.web.document.template.document_tag._jspx_meth_document_005fpdf_005f0(document_tag.java:289)
	at org.apache.jsp.tag.web.document.template.document_tag.doTag(document_tag.java:168)
	at org.apache.jsp.WEB_002dINF.views.responsive.documents.order.modelContract_jsp._jspx_meth_template_005fdocument_005f0(modelContract_jsp.java:127)
	at org.apache.jsp.WEB_002dINF.views.responsive.documents.order.modelContract_jsp._jspService(modelContract_jsp.java:92)
	... 147 more
Caused by: java.lang.NullPointerException
	at com.openhtmltopdf.render.InlineText.calcTotalAdjustment(InlineText.java:316)
	at com.openhtmltopdf.render.InlineLayoutBox.adjustHorizontalPosition(InlineLayoutBox.java:911)
	at com.openhtmltopdf.render.InlineLayoutBox.adjustHorizontalPosition(InlineLayoutBox.java:922)
	at com.openhtmltopdf.render.InlineLayoutBox.adjustHorizontalPosition(InlineLayoutBox.java:922)
	at com.openhtmltopdf.render.LineBox.adjustChildren(LineBox.java:270)
[...]

PR #421, opened 2019-11-29
handle NPE in justify text edge case #420

syjer:
hi @danfickle , this fix the issue #420 . From the debugging session, I see the issue is caused by a `InlineText`: "\n"  which I guess is generated by the `br` element, but I can't get a clear picture.

If I understood correctly, the align logic is calling `InlineText.calcTotalAdjustment` without calling `InlineText.countJustifiableChars` on that `InlineText` element before (which has the `InlineLayoutBox`: "br" as a parent), thus not creating a "_counts" instance.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/render/InlineText.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/render/InlineText.java
@@ -312,6 +312,10 @@ public class InlineText {
             // letter spacing is already explicitly set.
             return 0f;
         }
+
+        if (_counts == null) {
+            return 0f;
+        }
         
         return (_counts.getSpaceCount() * info.getSpaceAdjust()) +
                (_counts.getNonSpaceCount() * info.getNonSpaceAdjust());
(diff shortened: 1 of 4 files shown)


danfickle:
Hi @syjer,

I'm going to merge this for the test, but I'm not sure if the a better fix might be to go ahead and call `countJustifiableChars` if `_counts` is null. I'll have to debug to see why it is not being called though.

Thanks for the PR and sorry for the delay in merging.

syjer:
>call countJustifiableChars if _counts is null.

I was wondering the same. I think I tried to do that way, but I was missing a decent way to get the `CharCounts counts` parameter.

> sorry for the delay in merging.

no problem at all :)

danfickle:
It runs out zero is the correct result! The only time it doesn't count characters beforehand is for non-justified content nested inside justified text, so returning zero adjustment for the non-justified inline text is correct.

The reason it showed up on `<br/>` is that this element is implemented in the default stylesheet with `white-space: pre` which does not get justified.

The key method to figure this out was:
````java
    public void countJustifiableChars(CharCounts counts) {
        boolean justifyThis = getStyle().isTextJustify();
        for (Iterator<Object> i = getInlineChildren().iterator(); i.hasNext(); ) {
            Object o = i.next();
            if (o instanceof InlineLayoutBox) {
                ((InlineLayoutBox)o).countJustifiableChars(counts);
            } else if (o instanceof InlineText && justifyThis) {
                ((InlineText)o).countJustifiableChars(counts);
            }
        }
    }
````
Note: The `justifyThis` variable.

I added an extra test to make sure this was the case and it indeed breaks if we try counting chars when `_counts` is null.

Thanks again.

Issue #422, opened 2019-12-05
Problem with display cyrillic characters as #


MichaelZaleskovsky:
Hi! I have the problem with cyrillic characters with openhtmltopdf.
I want to get PDF from html, which contains cyrillic. I use this library in servlet.
But the cyrillic simbols converted to #

I use pdfBulider in this way:
pdfBuilder
.useFastMode()
.withHtmlContent(html, "/")
.toStream(stream)
.run();
and html string contains:

`<style> 
  @font-face { 
    font-family: "PT Sans"; src: url("https://fonts.googleapis.com/css?family=PT+Sans&subset=cyrillic"); 
  } 
  body { 
    font-family: 'PT Sans', Geneva, Arial, Helvetica, sans-serif; 
  } 
</style>`
But it doesn't help me.
Could you help me - what is the problem?

syjer:
hi @MichaelZaleskovsky ,

I don't think openhtmltopdf support the css generated by fonts.googleapis.com. You need to follow the guide https://github.com/danfickle/openhtmltopdf/wiki/Fonts#css-font-embedding-example .

So basically, you download your font (in TTF format) and define your "PT Sans" font as:

```css
@font-face {
font-family: 'PT Sans';
src: url(fonts/PTSans-Regular.ttf);
font-weight: normal;
font-style: normal;
}
```

the url is where your font is :)

Alternatively, you can register your font programmatically as described in https://github.com/danfickle/openhtmltopdf/wiki/Fonts#programatically-adding-fonts so you can avoid to define using `@font-face` .

https://fonts.google.com/specimen/PT+Sans?selection.family=PT+Sans -> you need to download the font, in the zip you will find `PTSans-Regular.ttf` which you can then use.

danfickle:
Assumed solved, reopen if you need to.

PR #423, opened 2019-12-05
In PrimitivePropertyBuilder.Src -> try to narrow down the PropertyValue list to a single value. This allow to link directly a google font css

syjer:
Hi @danfickle , looking at the issue #422 I was wondering why it skipped totally the `src` property in the following declaration:

```
@font-face {
  font-family: 'PT Sans';
  font-style: normal;
  font-weight: 400;
  src: local('PT Sans'), local('PTSans-Regular'), url(https://fonts.gstatic.com/s/ptsans/v11/jizaRExUiTo99u79D0KEwA.ttf) format('truetype');
}
```

It looks like `GenericURIWithNone` (which Src extends) accept only a list with a single element, thus  ignoring the src declaration.

Adding a logic to get the first matching element which is _not_ a unknown css type seems to be a decent heuristic.

As you can see from the test, you can now simply link a css from google fonts and it will work!

WDYT? :)

EDIT: by the way, I've noticed  `CSSParser` & co is not fully generified (lot of raw List parameters/types), are you interested in a small cleanup of that class? I think I can provide a PR :D


danfickle:
Hi @syjer,

Thanks for the effort on this. However, I haven't used it directly as I thought it could be made a bit more robust by checking the format tag after the uri.

Also, I'm trying to minimise network loads in the tests because my mac is getting slower and slower running them. On this note, could you please consider, when making a test using the `@page size` property to size the page as small as possible (to fit the test). This means less pixels to compare.

Thanks again,
Daniel.

syjer:
@danfickle  thank you for taking care of it :) I agree with your changes.

>On this note, could you please consider, when making a test using the @page size property to size the page as small as possible (to fit the test). This means less pixels to compare.

Sure, I didn't know about this issue at all 👍

Issue #428, opened 2020-01-14
Sources jar contains absolute local path


gabro:
The sources jar published on Maven contains an absolute local path to the licenses:

```bash
> jar -tf openhtmltopdf-pdfbox-0.0.1-RC19-sources.jar
META-INF/MANIFEST.MF
META-INF/
com/
com/openhtmltopdf/
com/openhtmltopdf/pdfboxout/
com/openhtmltopdf/pdfboxout/quads/
com/openhtmltopdf/pdfboxout/visualtester/
/
/Users/
/Users/daniel/
/Users/daniel/Documents/
/Users/daniel/Documents/workspace/
/Users/daniel/Documents/workspace/flyingsaucer/
/Users/daniel/Documents/workspace/flyingsaucer/target/
/Users/daniel/Documents/workspace/flyingsaucer/target/checkout/
/Users/daniel/Documents/workspace/flyingsaucer/target/checkout/openhtmltopdf-pdfbox/
/Users/daniel/Documents/workspace/flyingsaucer/target/checkout/openhtmltopdf-pdfbox/target/
/Users/daniel/Documents/workspace/flyingsaucer/target/checkout/openhtmltopdf-pdfbox/target/classes/
/Users/daniel/Documents/workspace/flyingsaucer/target/checkout/openhtmltopdf-pdfbox/target/classes/META-INF/
META-INF/maven/
META-INF/maven/com.openhtmltopdf/
META-INF/maven/com.openhtmltopdf/openhtmltopdf-pdfbox/
com/openhtmltopdf/pdfboxout/quads/KongAlgo.java
com/openhtmltopdf/pdfboxout/PdfBoxFontResolver.java
com/openhtmltopdf/pdfboxout/PdfBoxFastOutputDevice.java
com/openhtmltopdf/pdfboxout/PdfRendererBuilder.java
com/openhtmltopdf/pdfboxout/PdfBoxReplacedElementFactory.java
com/openhtmltopdf/pdfboxout/BookmarkElement.java
com/openhtmltopdf/pdfboxout/PdfBoxTextRenderer.java
/Users/daniel/Documents/workspace/flyingsaucer/target/checkout/openhtmltopdf-pdfbox/target/classes/META-INF/LICENSE-W3C-TEST
com/openhtmltopdf/pdfboxout/quads/Triangle.java
com/openhtmltopdf/pdfboxout/PdfBoxPerDocumentFormState.java
[...]

PR #430, opened 2020-01-15
fix issue #428: use relative path instead of ${project.build.outputDirectory}/META-INF

syjer:
This fix issue #428 . I've checked the resulting jars and they are now correctly built (the final jar, the source jar  the javadoc jars).

ATM I can't find any downside/issues.


danfickle:
Thanks mate. Merging now.

PR #450, opened 2020-03-02
Java2D - Ability to add fonts via code and manual test harness.

danfickle:
The manual tests can be run from:
`openhtmltopdf-examples\src\main\java\com\openhtmltopdf\testcases\j2d\Java2DVisualTest.java`

These tests are manual because Java2D rendering is not pixel identical across JREs, OSs, etc.

IF a test fails it will output an actual image, expected image and diff image at:
`working-directory\target\test\visual-tests\test-output\j2d`. The working directory will usually be the `openhtmltopdf-examples` module directory. If not there, check the parent directory as well.

From the diff image you will be able to work out if there is a bug in this library, or just a minor rendering difference by the JRE.

**NOTE:** Change to default behavior: Java2D will no longer pull in local machine fonts by default. To return to the previous behavior:
````java
builder.useEnvironmentFonts(true);
````

Otherwise, you should provide at least one font and use it on the body tag.

This change is to avoid differing behavior on development machine vs server.


ieugen:
I think this is a good move. Will make things reproducible and less dependent on environment. Looking forward to it.

Issue #406, opened 2019-11-12
PdfRendererBuilder.class.run throws Exception


jadams74:
This is a bad practice. 

https://github.com/danfickle/openhtmltopdf/blob/340f488d7b8ab37cf6b144eadae090f83428538b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfRendererBuilder.java#L36

It should throw throw render.createPDF()'s IOException instead.

leonorader:
Merging #460 should close this issue.

PR #460, opened 2020-03-21
changes Exception to IOException in PdfRendererBuilder #406

leonorader:


danfickle:
Thanks a lot @leonorader. Merging now.

Issue #469, opened 2020-04-27
a href mailto not working


mndzielski:
Hi @danfickle

Could you provide support for links with mailto?

````
<html>
<head>
<style>
</style>
</head>
<body>

user@gmail.com

<br /><br />

<a href="mailto:user@gmail.com">Send a message!</a>

</body>
</html>
````

I would like to customize text on email links.

syjer:
hi @mndzielski , you are right, the mailto: protocol is not supported, I've submitted a PR #470

PR #470, opened 2020-04-28
better handling of href uri #469

syjer:
Hi @danfickle , this fix the issue #469 . As the email URI was not matched by "://" I decided to directly use the URI.create route :).

I've tested with  http:// and mailto: manually and they both work now.


danfickle:
Thanks again @syjer!

Issue #477, opened 2020-05-12
Object Renderer link placement


hbergmey:
As outcome of #475 I am now rendering a tree graph using a custom object renderer based on Graphics2D. but I am now having extreme difficulties to getting the hyperlink boxes for the nodes in my graph correctly, which link to description pages in my document. By seeing where the mouspointer turns into a finger icon in Acrobat Reader I can see, that all link shapes seem to cover almost the whole region covered by the object. It does not matter, where I klick, it opens the same link, so this single link covers the whole region.

The display itself works fine and I am able to scale the graph to fit it in the page width. The graph consists of nodes connected by arrows. Each node is built from a Swing JPanel that lays out JLabels for Node ID, title, node information and outgoing edge ports using GridBagLayout. The nodes themselves are organized in a large JPanel using two nested GridBagLayouts. Again positioning works great and using SwingUtilities.convertRectangle I am able to determine the exact position relative to the main panel of each Title and NodeID label that is supposed to carry a link.

To collect the link shapes, I have extended the JLabel Component with an Interface that provides the url. So during rendering in the custom object drawer, I am able to collect all linked components after layout completes, calculate their bounding shapes relative to the Graphics 2D and create the link map, with shapes as keys and urls as values.

I have verified that the positions of the collected shapes match the drawn Ids and titles by drawing them as red boxes at the end of graph drawing. The cover exactly the areas that are supposed to be the hotspots of the links. But if I return the shape-to-url map and let opengtmltopdf even render just the firdt link, its hotspot covers almost the whole page, instead of the title it is supposed to cover.

I have tried several corrective transformations, even transforms I would expect to result in an extremely small region in the upper left corner on the page, but it does not change anything. It is still the whole region reacting to the click.

Since it does not seem to matter, how large or small I transform the collected link shapes, I think this is a bug and I do not have a clue, how to get it right.

hbergmey:
It appears to me, as if the complete object-Element becomes linked, instead of the individual shapes rendered by it's object renderer. For the following screenshot I have placed the mouse pointer on top of the graph object.

![grafik](https://user-images.githubusercontent.com/36409310/81672295-3d96b880-944a-11ea-998d-66f5797c3b66.png)
It is scaled extremely small to obfuscate the proprietary info. Sorry for that, but I think, what I would like to show, still becomes apparent.

[...]

PR #480, opened 2020-05-13
Fix link shapes #477

rototor:
Only use the scaling from the given transform

For whatever reason we only want the DPI scaling from the given transform, and
not any other property of the transform.

When adding a link to process later, the transform has at that moment also a translate in it. 

Before 1ffd27175926d8d25294500f71cc39f651421fc8 the used transform was after the last page was 
processed. At that moment the transform only had a scale component and no translate.


danfickle:
Hi @rototor, sorry for breaking this. After merging this pr, I'll try to create a basic test in non visual regression test so I don't break it again.

Thanks again.

Issue #482, opened 2020-05-14
OutOfMemoryError for HTML with long line(s)


swarl:
Good evening

Version 1.0.2
The following HTML sent to my application produces a hell of an OutOfMemoryError: 
```
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Strict//EN" "http://www.w3.org/TR/xhtml1/DTD/xhtml1-strict.dtd">
[...]

PR #483, opened 2020-05-15
#482: Smoke Testcase for the endless loop because of white-space hand…

rototor:
…ling.

It demonstrates the bug in #482, a endless loop while trying to break a line. Should be included to detect regressions with that in future.


rototor:
This is no longer a smoke test, but also provides a fix for this problem which seems to work and not break any other test.

syjer:
hi @rototor , I may be wrong, but I think it will still fail if the string is larger than 1 but contains only SOFT_HYPHEN characters.

I would guess for each input string, we should reduce it to always have only one consecutive SOFT_HYPHEN .

rototor:
@syjer You are right, I extended the testcase and the fix to also handle that cases.

danfickle:
Thanks both of you. That would have taken me ages to find.

Issue #484, opened 2020-05-15
 Base 64 encoded SVG does not load in pdf


adrianrodfer:
Hi,

I have an image with a Base 64 encoded "svg" in my html file, like this:

`<img src="data:image/svg+xml;base64,PD94bWwgdmVyc2...">`

 I need to have the image in a pdf. But I get the following error:

```
com.openhtmltopdf.load WARNING:: Unrecognized image format for: data:image/svg+xml;base64,PD94bWwgdmV...
com.openhtmltopdf.exception WARNING:: Can't read XHTML embedded image.
```

Is there any way to insert a Base64 encoded svg?

Thanks,
Adrian

syjer:
hi @adrianrodfer , at the moment it's not supported. But I'll provide a pull request asap.

rekhubs:
> hi @adrianrodfer , at the moment it's not supported. But I'll provide a pull request asap.

I need this too.... Thank you!

vasilyevia:
Hi @syjer. Is it possible to support plain text svg data like the following?
`data:image/svg+xml;utf8,<svg xmlns="http://www.w3.org/2000/svg" width="9" height="12" viewBox="0 0 9 12"><path ...`

syjer:
hi @vasilyevia , it should not be hard, could you open a new issue as a feature request so I can track it? Thank you.

vasilyevia:
> 
> 
> hi @vasilyevia , it should not be hard, could you open a new issue as a feature request so I can track it? Thank you.

Thanks. Created a new issue. Not sure how to mark it as a feature request.
https://github.com/danfickle/openhtmltopdf/issues/583

Abhinandan578:
I am still facing the same issue. SAVG Images are not getting loaded. I am using docx4j-ImportXHTML (11.4.8) which is calling openhtmltopdf-core(1.0.10). Can someone please help here.

jeong-donghee:
I resolved it by using a workaround of converting the SVG to PNG.

PR #485, opened 2020-05-15
implement support for data:image/svg+xml;base64 in img src. See issue #484

syjer:
Hi @danfickle , this PR implement the support of svg as data uri encoded in base64.


danfickle:
Thanks @syjer, perfect pr with test! Much appreciated.

Issue #444, opened 2020-02-12
How to provide resolver for SVG? com.openhtmltopdf.extend.FSUriResolver not being used by SVG 


ieugen:
Hello,

I'm having some difficulties rendering PDF sprites. I followed the instructions in https://github.com/danfickle/openhtmltopdf/wiki/Plugins:-SVG-Images#svg-sprites .

I expected the SVG code to use the logic in `FSUriResolver` implementation but it does not and I don't know how to provide an implementation to fix the path. 

I believe we should be able to change the base path for resolving SVG's in the filesystem / remote. 
Any hints as to where/how can this be done? 

Thanks,

More details bellow:

We use relative urls for the sprites and we get:
```
2020-02-12 22:19:21.670  WARN 520661 --- [ault-executor-0] com.openhtmltopdf.exception              : Couldn't draw SVG.

org.apache.batik.transcoder.TranscoderException: null
Enclosed Exception:
report-assets/sprites/solid.svg (No such file or directory)
	at org.apache.batik.transcoder.SVGAbstractTranscoder.transcode(SVGAbstractTranscoder.java:226) ~[batik-transcoder-1.12.jar:na]
	at com.openhtmltopdf.svgsupport.PDFTranscoder.transcode(PDFTranscoder.java:216) ~[openhtmltopdf-svg-support-1.0.1.jar:na]
	at org.apache.batik.transcoder.XMLAbstractTranscoder.transcode(XMLAbstractTranscoder.java:142) ~[batik-transcoder-1.12.jar:na]
	at org.apache.batik.transcoder.SVGAbstractTranscoder.transcode(SVGAbstractTranscoder.java:156) ~[batik-transcoder-1.12.jar:na]
	at com.openhtmltopdf.svgsupport.BatikSVGImage.drawSVG(BatikSVGImage.java:185) ~[openhtmltopdf-svg-support-1.0.1.jar:na]
	at com.openhtmltopdf.pdfboxout.PdfBoxSVGReplacedElement.paint(PdfBoxSVGReplacedElement.java:70) [openhtmltopdf-pdfbox-1.0.1.jar:na]
	at com.openhtmltopdf.pdfboxout.PdfBoxFastOutputDevice.paintReplacedElement(PdfBoxFastOutputDevice.java:275) [openhtmltopdf-pdfbox-1.0.1.jar:na]
[...]

PR #486, opened 2020-05-16
svg backend: use openhtmltopdf UserAgentCallback for fetching external resources #444

syjer:
Hi @danfickle , this PR is a work in progress for making more coherent the loading of resources in the svg backend (see issue #444)

Basically we inject our own `UserAgentCallback` which contain all the logic inside a custom `DocumentLoader`.

There still some open point:

 - currently it will block by default the loading through the `UserAgentCallback` as `allowExternalResources` is set false by default. Unfortunately at the moment it's a binary choice. We can totally open or totally close the access to the external world. How can we improve that so we can, for example, only allow some specific protocols (most likely a custom one)? Maybe we can extend the interface `UserAgentCallback` so he can provide some - customizable by the user - logic for checking the load of external resources ? Or maybe we pass a whitelist of protocol in the svg builder?

- tests :)

What do you think? :)

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/SVGDrawer.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/SVGDrawer.java
@@ -12,18 +12,20 @@ import com.openhtmltopdf.render.Box;
 import com.openhtmltopdf.render.RenderingContext;
 
 public interface SVGDrawer extends Closeable {
-    public void importFontFaceRules(List<FontFaceRule> fontFaces,
+    void importFontFaceRules(List<FontFaceRule> fontFaces,
             SharedContext shared);
 
-    public SVGImage buildSVGImage(Element svgElement, Box box, CssContext cssContext, double cssWidth,
+    SVGImage buildSVGImage(Element svgElement, Box box, CssContext cssContext, double cssWidth,
             double cssHeight, double dotsPerPixel);
 
-    public static interface SVGImage {
-        public int getIntrinsicWidth();
+    default void withUserAgent(UserAgentCallback userAgentCallback) {}
 
-        public int getIntrinsicHeight();
+    interface SVGImage {
+        int getIntrinsicWidth();
 
-        public void drawSVG(OutputDevice outputDevice, RenderingContext ctx,
+        int getIntrinsicHeight();
+
+        void drawSVG(OutputDevice outputDevice, RenderingContext ctx,
                 double x, double y);
     }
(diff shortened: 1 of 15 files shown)


danfickle:
Hi @syjer,

The way I've handled this in general is everything is meant to go through any configured uri resolver which can return null if it doesn't want to allow a particular resource. Uri resolver with stream handlers means that the user should have total customisation options to loading resources. The move to disallow external resources in svg by default was because it was not using the uri resolver.

Ideally the svg renderer would use the configured uri resolver and any stream handlers. We may then not need the ability to provide the custom user agent? If you have found where to plugin to batik to intercept resource calls, we could just call into our user agent which is already aware of uri resolver and stream handlers.

What do you think? Thanks again.

syjer:
hi @danfickle , 

> Ideally the svg renderer would use the configured uri resolver and any stream handlers. We may then not need the ability to provide the custom user agent? If you have found where to plugin to batik to intercept resource calls, we could just call into our user agent which is already aware of uri resolver and stream handlers.

This is exactly what this PR is doing :)

The most important line is: https://github.com/danfickle/openhtmltopdf/pull/486/files#diff-ab8a6ba6ef10e59d4db15a47c794c89dR23

. Everything else is for making it possible to inject our user agent inside Batik.

I'll begin to add tests asap.

danfickle:
That's embarrassing! I should have read the code more carefully. Yes, with tests this pr would be awesome. As to the question of whether we should allow external resources by default, albeit through the uri resolver, etc, I think we're stuck with the current situation. If they want external resources, use the other constructor. Otherwise we would be silently allowing external resources. Not everybody reads the change log!

danfickle:
Hi @syjer,

Unless you or @rototor have any objections, I plan to do a release tomorrow or Monday. If you have no objections, I'll merge this beforehand and add a couple of quick tests.

Thanks again.

syjer:
hi @danfickle , no objection on my side. I'll add some tests today :)

rototor:
@danfickle I also have no objection.

syjer:
@danfickle , I've added the possibility in the `BatikSVGDrawer` constructor to specify a whitelist of protocols.

Now I'll add the tests :)

syjer:
@danfickle , I've added 3 visual tests:

- one that check if it will block the loading of external resource (default)
- one that whitelist the file protocol and check the case of relative file url
- one with a custom:/ protocol

danfickle:
Awesome, you've even used `@page` sizes to minimise test runtime! Thanks!

Issue #76, opened 2017-03-07
Replace XRLogger with some logging framework API


aleksandr-m:
Log4j2 perhaps? And get rid of `openhtmltopdf-log4j` and `openhtmltopdf-slf4j`.
Wdyt?

danfickle:
G'day @aleksandr-m 
I was thinking of the following use case. Imagine a CMS to output PDF documents. There are many templates and an on-line template editor to create more. Currently, the only way for the template author to see her errors is to have access to the server logs. Even if she has access, the output for the template she is editing is mixed in with output for all the other templates that may be output at the same time. Wouldn't it be better if she could see the errors for the template she is working on in the online editor?

To support this use case I propose to have a log interceptor for each run configurable in the builder. Then the log output could be shown in the online editor. While doing this I suggest we also move to logging a enum constant (with associated message format) rather than a string directly for anyone that wants to translate or filter log messages. So the log message class might look like following:

````java
public class LogMessage {
   private final LogMessageId id;
   private final Object[] args;
   private final Level level;

   // constructor...
   // getters...
}
````
and the ````LogMessageId```` enum like following:

````java
public enum LogMessageId {
  UNRECOGNIZED_CSS_PROPERTY("The CSS property {} is not recognized"),
  UNIMPLEMENTED_CSS_VALUE("The CSS value {} is not currently implemented");

  private final String msgFormat;
  private LogMessageId(String format) {
    this.msgFormat = format;
  }

  public String getMessageFormat() {
    return msgFormat;
  }
}
````

Then we would provide a ````DefaultStringLogInterceptor```` something like:
````java
public class DefaultStringLogInterceptor implements LogReceiver {
  private final StringBuilder sb = new StringBuilder();

  @Override
  public void receive(LogMessage msg) {
    sb.append(MessageFormat.format(msg.getId().getMessageFormat(), msg.getArguments()));
  }
  
  @Override
  public String toString() {
    return sb.toString();
  }
}
````

As for Log4j2, I would be wary of including it directly as it may cause jar version hell for many users of the library. I agree we should get rid of ````openhtmltopdf-log4j```` and ````openhtmltopdf-slf4j```` and replace with examples in the documentation (they're both single files) because I doubt anyone is using them because of aforementioned version hell.

OK, that's a long post, anyway, what do you think?

aleksandr-m:
[...]

PR #489, opened 2020-05-25
Add possibility to fetch all logged warning programmatically, see #76

syjer:
Hi @danfickle , this PR is still a work in progress, but the main functionality should be here.

As you described in the issue #76 (https://github.com/danfickle/openhtmltopdf/issues/76#issuecomment-284936563) , adding a way to fetch all logged values programmatically is quite a nice feature, this PR enable that.

I still need to do:

 - check if the cleanup is right in all cases
 - expose the Diagnostic values (add getters)
 - maybe add an option to enable the recording of log in the builder, instead of doing it by default
 - tests

What do you think ? :)

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/ThreadCtx.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/ThreadCtx.java
@@ -2,27 +2,42 @@ package com.openhtmltopdf.util;
 
 import com.openhtmltopdf.layout.SharedContext;
 
+import java.io.Closeable;
+import java.util.function.Consumer;
+import java.util.logging.Level;
+
 /**
  * Because OpenHTMLtoPDF is designed to run in a single thread at all times for one invocation,
  * we can use a ThreadLocal to store pseudo global variables.
  * This MUST be set up in the appropriate renderer.
  */
 public class ThreadCtx {
-	private static final ThreadLocal<ThreadData> data = new ThreadLocal<ThreadCtx.ThreadData>() {
-		@Override
-		protected ThreadData initialValue() {
-			return new ThreadData();
-		}
-	};
-	
+
+	private static final ThreadLocal<ThreadData> data = ThreadLocal.withInitial(() -> new ThreadData());
+	private static final ThreadLocal<Consumer<Diagnostic>> diagnosticConsumer = new ThreadLocal();
+
+
+
 	public static ThreadData get() {
 		return data.get();
+	}
+
+	static void addDiagnostic(Diagnostic diagnostic) {
+		Consumer<Diagnostic> consumer = diagnosticConsumer.get();
+		if (consumer != null) {
+			consumer.accept(diagnostic);
+		}
 	}
 	
 	public static void cleanup() {
 		data.remove();
 	}
-	
+
+	public static Closeable applyDiagnosticConsumer(Consumer<Diagnostic> consumer) {
+		diagnosticConsumer.set(consumer);
+		return () -> diagnosticConsumer.remove();
+	}
+
 	public static class ThreadData {
 		private ThreadData() { }
 		private SharedContext sharedContext;
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/XRLog.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/XRLog.java
@@ -33,7 +33,7 @@ import java.util.logging.Level;
  * @author empty
  */
 public class XRLog {
-    private static final List<String> LOGGER_NAMES = new ArrayList<String>(20);
+    private static final List<String> LOGGER_NAMES = new ArrayList<>(20);
     public final static String CONFIG = registerLoggerByName("com.openhtmltopdf.config");
     public final static String EXCEPTION = registerLoggerByName("com.openhtmltopdf.exception");
     public final static String GENERAL = registerLoggerByName("com.openhtmltopdf.general");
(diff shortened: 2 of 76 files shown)


danfickle:
Hello @syjer,

I was just thinking, why not allow the user to provide a `Consumer` in the builder? That way they can do whatever they like with them rather than just an `ArrayList`.

Eventually, I think of two things with "logging":
1. Log with enum constants (convertible to strings), so the user can programmatically determine what is being logged without resorting to ugly string comparison.
2. The ability to throw a poison pill exception on any selected log callback. This would allow for example some users to stop the rendering on any load failure while others would want a more tolerant approach. Currently the user can not configure what happens.

Of course, 1 is a lot of work!

syjer:
hi @danfickle 

>I was just thinking, why not allow the user to provide a Consumer in the builder? That way they can do whatever they like with them rather than just an ArrayList.

You are right! Nice idea, it will simplify some logic too.

 
1: yep, I'll see how many different logging messages are present. It may be a bit mechanical as a work, but I guess it will pay off.

2: agree, I guess using the builder approach + enum, we can ensure it will be possible to let the user block the execution as he want.

I'll try to update the code for next week :)

syjer:
switched to the Consumer approach.

Next I'll try to switch to the enum variant

openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/XRLog.java
```diff
@@ -66,176 +66,218 @@ private static String registerLoggerByName(final String loggerName) {
      */
     public static List<String> listRegisteredLoggers() {
         // defensive copy
-        return new ArrayList<String>(LOGGER_NAMES);
+        return new ArrayList<>(LOGGER_NAMES);
     }
 
+    public enum LogMessageId {
```
danfickle: I think `LogMessageId` will be big enough to warrant its own file.

Also, I'm worried that since we have a lot of code paths that are not tested, that we will get into trouble with mismatching argument counts for the message format. One way to alleviate this problem would be to have an message enum for each arg count and a corresponding method which accepted that enum and the correct number of args. Then we would make the enums implement a common interface, which is what the user would see. Interested in your feedback.

openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/ThreadCtx.java
```diff
@@ -2,26 +2,64 @@
...
+
+	public static void addDiagnostic(String where, Level level, String msg, Throwable th) {
+		Consumer<Diagnostic> consumer = diagnosticConsumer.get();
+		if (consumer != null) {
+			consumer.accept(new Diagnostic(where, level, msg, th));
+		}
+	}
 	
 	public static void cleanup() {
 		data.remove();
```
danfickle, line 33: I think we have to `diagnosticConsumer.remove()` here to avoid zombie data.

openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/ThreadCtx.java
```diff
@@ -2,26 +2,64 @@
...
+	public static class Diagnostic {
+		private final String where;
+		private final Level level;
+		private final String msg;
+		private final Throwable error;
+
+		public Diagnostic(String where, Level level, String msg, Throwable error) {
+			this.where = where;
+			this.level = level;
+			this.msg = msg;
```
danfickle: This implies that the user gets the message as string, wouldn't it be better to pass message enum and arg array in case they don't want to pay the price of formatting?

openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/ThreadCtx.java
danfickle: I see on second thoughts that you have dealt with it using closeable of your own in apply method. Your method is probably cleaner.

openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/ThreadCtx.java
danfickle: Example: They are not interested in info level messages.

danfickle:
I've added a few inline comments, but I'm generally pleased with the look of this. My thinking was very similar.

openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/ThreadCtx.java
syjer: you are right, it's something that I need to update. I'll switch to the enum+params.

openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/XRLog.java
syjer: agree about the new file. 

About the mismatching count of parameters/message, I agree we should be able to statically prevent the misuse. I'm wondering if there is an alternative way to implement it, losing the single enum type is unfortunate.

syjer: done, I think it will work :)

syjer:
I've pushed some additional code following your feedback, I think it's beginning to look decent :).

Btw I've added a `FIXME` comment here https://github.com/danfickle/openhtmltopdf/pull/489/files#diff-1240d78636f3f4023a60d044cd8184eaR97 , as I'm still not sure how to bridge the new `Diagnostic` format to the logger interface in an elegant way.

danfickle:
You have been busy! In regards to the legacy logging situation, I think the core problem is that `java.util.logging` where the log messages end up by default, does not support parameters, just a single string, which is a performance waste if a log message is not ultimately used. Fortunately, slf4j , which most people probably already have on the classpath, does support parameters. 

Therefore, I'm thinking that we work on the slf4j logger module (to use the new parameterised log messages, will require a new interface) and recommend that people use it. Of course, users can continue to use `j.u.l` at a performance penalty.

P.s. if we stick with an old version of slf4j, users can exclude our copy via maven, as it is supposedly binary backwards compatible.

syjer:
yep. I'm organizing the code to have 0 overhead for slf4j.

I'm hoping to complete the PR this week :)

syjer:
I've pushed the diagnostic bridging implementation:

A default one: https://github.com/danfickle/openhtmltopdf/pull/489/commits/69fb5f558c851484508b1af13883d22db2c03eff#diff-14a42331e2299f386db6d8cc423d57c8R39

And a custom one for slf4j:

https://github.com/danfickle/openhtmltopdf/pull/489/commits/69fb5f558c851484508b1af13883d22db2c03eff#diff-6da5c18b096d790496d094f126d941f1R58

syjer:
I think I got everything now :).

syjer:
after the merge of #507 I need to resolve some conflicts, I'll do it today :)

danfickle:
One thing I've just noticed, the `buildpdfrenderer` method in `pdfrendererbuilder` is public, and used by users wanting the `pddocument` for post processing. Therefore, it needs to behave the same as run for the consumer.

Other than that it all looks excellent online, but I will have a look on my ide after you fix merge issues.

P.s. I appreciate the massive amount of work in this pr, so another big thanks.

syjer:
>One thing I've just noticed, the buildpdfrenderer method in pdfrendererbuilder is public, and used by users wanting the pddocument for post processing. Therefore, it needs to behave the same as run for the consumer.

I guess I forgot to cover that case. I'll update there too.

>P.s. I appreciate the massive amount of work in this pr, so another big thanks.

You're welcome :)

syjer:
In the last commit https://github.com/danfickle/openhtmltopdf/pull/489/commits/e705452267e2407778b28a344e430f74df918675 I've propagated the `Consumer` inside the `*Renderer` and wired up the closing of the consumer inside their `close()` method.

With that I think we have covered all the cases?

danfickle:
Yes, I think that is everything. Thanks again.

syjer:
thank you for the review and merging 👍 .

I'll most likely do another round of cleanup as I think there is still some old unused code and pattern that can be improved using java8 api.

PR #491, opened 2020-05-26
add failing test that highlight the infinite loop issue in the inline box layout algorithm #482

syjer:
Hi @danfickle , I've added another failing test that cause an infinite loop somewhere in ~~the table layout algorithm.~~ the inline box layout algorithm

I've finally managed to reduce it in a more minimal case. As described in https://github.com/danfickle/openhtmltopdf/issues/482#issuecomment-633999078 there are multiple elements in play here:

 - the fonts seems to have an impact
 - it seems to be still a `SOFT_HYPHEN` issue
 - the defined width of the `td` element seems to have an impact too

I'll try to understand what is true culprit :)

### DIFF
--- a/openhtmltopdf-examples/src/main/resources/visualtest/html/issue-482-infinite-loop-table.html
+++ b/openhtmltopdf-examples/src/main/resources/visualtest/html/issue-482-infinite-loop-table.html
@@ -1,0 +1,19 @@
+<html>
+<head>
+<style>
+    * {
+        font-family: 'Liberation Sans', sans-serif;
+        margin: 0;
+        padding: 0;
+    }
+
+    .content {
+        word-wrap: break-word;
+        white-space:pre-wrap;
+    }
+</style>
+</head>
+<body>
+<table class="content"><tbody><tr><td width="10">­</td></tr></tbody></table>
+</body>
+</html>
--- a/openhtmltopdf-examples/src/test/java/com/openhtmltopdf/visualregressiontests/VisualRegressionTest.java
+++ b/openhtmltopdf-examples/src/test/java/com/openhtmltopdf/visualregressiontests/VisualRegressionTest.java
@@ -1154,6 +1154,15 @@ public class VisualRegressionTest {
         })));
     }
 
+    @Test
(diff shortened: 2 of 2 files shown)


syjer:
Additional info, it's still stuck inside https://github.com/danfickle/openhtmltopdf/blob/open-dev-v1/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/InlineBoxing.java#L160

It seems that the `SOFT_HYPHEN` character with the `Liberation Sans` font  has a length which is more than 0.

danfickle:
Thanks @syjer, I'll merge and have a look right away.

syjer:
hi @danfickle , I've narrowed down a little bit more :

the final file is 

```
<html>
<head>
<style>
@page  {
    size: 30px 30px;
    margin: 0;
    padding:0;
}
    .content {
        font-family: 'Liberation Sans', sans-serif;
        word-wrap: break-word;
        width:10px;
    }
</style>
</head>
<body>
<div class="content">­</div>
</body>
</html>
```

syjer:
the final file is 

```
<html>
<head>
<style>
@page  {
    size: 30px 30px;
    margin: 0;
    padding:0;
}
    .content {
        font-family: 'Liberation Sans', sans-serif;
        word-wrap: break-word;
        width:10px;
    }
</style>
</head>
<body>
<div class="content">­</div>
</body>
</html>
```

danfickle:
I've committed a fix but I still don't think it is robust enough. The problem is we have pseudo code like this:
````
while (reported width less than line width) {
   rerun word breaking algorithm
}
````
So that if the word breaking algorithm measures and determines it's impossible but under reports the width an infinite loop bug will occur. This was happening because the reported width was about 6px but the measured width with the inserted hyphen was twice this.

A better solution would be to go ahead and output the rest on a new line if the first run reports a width less than line width (ie. DRY).

I'll try this tomorrow. In the meantime, I'll add a comment in the release issue about `break-word` not being suitable for production.

P.S @syjer you were spot on about the width of a soft-hyphen being a problem. Thanks.

PR #492, opened 2020-05-27
#482 #483 #491 Make word breaker testable and start writing tests.

danfickle:
We need a much more robust line breaker so any suggested unit tests will be incorporated.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/Breaker.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/Breaker.java
@@ -446,7 +446,8 @@ public class Breaker {
                         extraSpacing + letterSpacing);
                 current.withHyphenGraphicsLength = current.graphicsLength + withTrailingHyphenSplitWidth;
                 
-                if (current.withHyphenGraphicsLength > avail) {
+                if (current.withHyphenGraphicsLength >= avail &&
+                    current.right != currentString.length()) {
                     current.graphicsLength = current.withHyphenGraphicsLength;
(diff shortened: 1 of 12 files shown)


syjer:
I wonder if we could do a property based test in this case (using something like [junit quickcheck](https://github.com/pholser/junit-quickcheck/) ).

danfickle:
This pr does the following:
+ Makes line breaker (by word) unit testable.
+ Adds several unit tests to exercise line breaker.
+ Adds end to end random fuzz testing for the line breaking. This includes all `white-space` values, both `break-word` and normal line breaking, with fonts which have a zero width soft hyphen as well as one that does not, in divs from 0px wide upwards. A random sampling of 5 character sequences (with the five characters chosen being special meaning in the line break algorithms) is used when the test is run each time, with the seed output to logs so any troublesome runs can be investigated. Hopefully, this random and regular testing will stop endless loop bugs creeping in again.
+ A safety valve for line breaking attempts when `break-word` is in effect. Triggering this safety check will deliberately result in a fatal exception which should be reported. Yes, this is a code smell, but one thought necessary after at least 3 endless loop bugs in my implementation of this complex functionality. We should be able to remove the safety check if the current fixed implementation proves robust.

danfickle:
> I wonder if we could do a property based test in this case (using something like [junit quickcheck](https://github.com/pholser/junit-quickcheck/) ).

I initially thought of parameter testing, but was underwhelmed by the syntax in Junit 4. After looking at the documentation of quickcheck, I probably should have used it rather than reinventing the wheel...

PR #496, opened 2020-05-30
Adding ability to specify 'UserStyleSheetURI' for an SVGDrawer, this …

### LINKED ISSUE #493: External css file for SVG styles- best practices?
Is there support for applying styles from an external css file to inline svg's?  I've done some tinkering around adding processing instructions to the svg and transcoding hints without much luck.

amckain92:
…allows SVGs to render without requiring styles to be inline.
Fixes #493

Hi  @danfickle,

So I was trying to find a solution to issue 493 where in order for SVG elements to render with the correct css, the styles are required to be declared inline in the SVG.  Ideally, SVG elements would be able to reference an external stylesheet to pull their styles as well.  This was possible in the older itext/flying saucer implementation by providing a transcoder hint for the svg/png transcoder used by Batik.

Eventually I tracked down the issue in the Batik BridgeContext class, specifically the 'initializeDocument' method tries to set up the CSSEngine on the SVGDOMImplementation using the UserAgent.

The problem that I was running into was the implementation of UserAgent, OpenHtmlUserAgent doesn't override the method 'getUserStyleSheetURI' from its base class UserAgentAdapter and thus always returns null for that method call.  That method is what's used in the BridgeContext to setup the CSSEngine for the SVG.

To get the value I needed injected down to the UserAgent class I had to pass it through a number of other classes starting at the public interface point which is the BatikSVGDrawer:
BatikSVGDrawer > BatikSVGImage > PDFTranscoder > OpenHtmlUserAgent

I'm not super familiar with this library, so there may have been a more elegant implementation path that I missed, also unsure of any potential side affects this may have caused but I tried to be as minimal and backwards compatible as possible.  Only obvious breaking change I can see is the modification of the existing constructor in OpenHtmlUserAgent which could be breaking if that class is used outside the library, but that could be alleviated by adding the new constructor along side the existing constructor.  Using the new stylesheet property would also require the user to construct the BatikSVGDrawer with the argument SvgExternalResourceMode.INSECURE_ALLOW_EXTERNAL_RESOURCE_REQUESTS which may not be ideal.

Example usage:

```
BufferedOutputStream outputStream = null;
        try {
            outputStream = new BufferedOutputStream(new FileOutputStream(output));
            PdfRendererBuilder builder = new PdfRendererBuilder();
            builder.useFastMode();
            builder.withHtmlContent(html,executingUri);
            builder.toStream(outputStream);

            BatikSVGDrawer svgDrawer = new BatikSVGDrawer(BatikSVGDrawer.SvgScriptMode.SECURE, 
     BatikSVGDrawer.SvgExternalResourceMode.INSECURE_ALLOW_EXTERNAL_RESOURCE_REQUESTS);
            
            //new method added to inject external style sheet for SVG            
            svgDrawer.withUserStyleSheetURI("classpath://SOME_PATH/svgStyles.css");
            builder.useSVGDrawer(svgDrawer);
            builder.run();

        } catch (Exception e) {
            logger.error("Exception rendering pdf", e);
        }
        finally {
            outputStream.close();
        }
```

Thank you for the review and consideration!

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/SVGDrawer.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/extend/SVGDrawer.java
@@ -20,6 +20,8 @@ public interface SVGDrawer extends Closeable {
 
     default void withUserAgent(UserAgentCallback userAgentCallback) {}
 
+    default void withUserStyleSheetURI(String userStyleSheetURI) {}
+
     interface SVGImage {
         int getIntrinsicWidth();
 
--- a/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/BatikSVGDrawer.java
+++ b/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/BatikSVGDrawer.java
@@ -22,6 +22,7 @@ public class BatikSVGDrawer implements SVGDrawer {
     private final boolean allowScripts;
     private final boolean allowExternalResources;
     private UserAgentCallback userAgentCallback;
+    private String userStyleSheetURI;
     
     public enum SvgScriptMode {
         SECURE,
@@ -84,6 +85,9 @@ public class BatikSVGDrawer implements SVGDrawer {
     }
 
     @Override
+    public void withUserStyleSheetURI(String userStyleSheetURI) { this.userStyleSheetURI = userStyleSheetURI; }
+
+    @Override
     public SVGImage buildSVGImage(Element svgElement, Box box, CssContext c,
     		double cssWidth, double cssHeight, double dotsPerPixel) {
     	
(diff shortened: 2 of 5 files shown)


syjer:
hi @amckain92 , personally I would prefer if the user don't need to specify a specific css for the svg, but instead, like a normal browser, it would apply the defined styles currently present.

That's why I'm kinda exploring the possibility to modify the dom of the svg elements and inline the css affecting them. I think it would be more transparent and less magic.

Clearly, YMMV and your solution is a nice stop gap :)

danfickle:
I will look at this, specifically, how hard it is to extract styles to string once matches/cascade has run.

danfickle:
Hi @amckain92,

I have now implemented auto-extraction of SVG styles as suggested by @syjer in #515. Maybe, you could have a look at it and see if it fits your requirements.

By the way, big thanks for contributing to the project.

danfickle:
Ok, closing now. Thanks again for this contribution.

PR #507, opened 2020-06-23
remove more unused utils/misc cleanup

syjer:
hi, this PR remove/reorganize even more unused utils code.


danfickle:
Thank you. It turns out I recreated the permutation generator in #492. It took me a while too! Anyway, I'm trying to get back into this project, so thanks for keeping it going.

Issue #508, opened 2020-07-07
PDAnnotationFileAttachment support


lindamarieb:
Is there any way to indicate that an anchor is a link to an attached file?

I can add annotations manually in my Java code to the attachment using PDAnnotationFileAttachment but I need to place the links in my html.

syjer:
Hi @lindamarieb , as far as I know, currently there is no support for this feature, unfortunately.

I'll have a look and try to provide a pull request.

syjer:
I did a first prototype, it's a little bit viewer dependent:

(note: I've been able to hide the icon in Evince, sumatrapdf and Adobe reader)

 - embedded pdf viewer from chrome and mac os: does not work, you see the text ("test") but it does nothing on click
 - evince (pdf viewer from gnome, linux), it works, but as you can see the text is hidden by the icon: 
  ![Screenshot from 2020-07-08 14-53-15](https://user-images.githubusercontent.com/498146/86921196-06cbde80-c12b-11ea-8e1b-36fca76c8805.png)
 - embedded viewer in edge/ie (windows):
   ![Capture_png](https://user-images.githubusercontent.com/498146/86921267-2400ad00-c12b-11ea-9749-9b08424fe6e7.PNG)
 - sumatrapdf (windows):
   ![Capture2_png](https://user-images.githubusercontent.com/498146/86921349-48f52000-c12b-11ea-8a20-531b3bc05106.PNG)
 - adobe:
   <img width="406" alt="Capture_3" src="https://user-images.githubusercontent.com/498146/86921668-c6b92b80-c12b-11ea-8da1-1ad2517a6eca.png"> 


It does not seems possible to hide the icon as far as I know. It's a little bit ugly unfortunately.

syjer:
code: https://github.com/syjer/openhtmltopdf/commit/cdb50764827c1001a22354ef72a933a528066589

example of input file:

```
<html>
<head>
<style>
</style>
</head>
<body>
<a href="https://user-images.githubusercontent.com/498146/86921196-06cbde80-c12b-11ea-8e1b-36fca76c8805.png" data-embed-file="true" data-content-type="application/octet-stream">Test</a>
</body>
</html>
```

the presence of `"data-embed-file"` attribute on a `a` element, will trigger the download of the linked resource (could be a local file, or a custom protocol).

syjer:
note: I've been able to find a way to generally hide the icon. Only the embedded viewer from ie/edge persist to show it.

[...]

PR #509, opened 2020-07-08
#508 add support for embedded file

syjer:
hi, this PR implement the support of embedding files in the pdf.

You can see in https://github.com/danfickle/openhtmltopdf/issues/508#issuecomment-655509628 how it work. Basically, adding the attribute `data-embed-file="true"` to a link will make the linked resource embedded.

As written in https://github.com/danfickle/openhtmltopdf/issues/508#issuecomment-655504301 , the support from the pdf viewer is, uh, variable. ~~Especially the icon is quite ugly, and at the moment I'm not able to hide it.~~ Edit: I was able to customize it. Seems to globally work, except for the viewer from IE/Edge. But at least adobe reader seems to work.

Note: there are security implication if a third party user has control over the html. Maybe we need to add an explicit list of embeddable uri resource in the builder?

I still need to add some kind of test.

WDYT?

### DIFF
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
@@ -253,7 +253,8 @@ public class PdfBoxFastLinkManager {
 					try {
 						PDComplexFileSpecification fs = new PDComplexFileSpecification();
 						PDEmbeddedFile embeddedFile = new PDEmbeddedFile(_od.getWriter(), new ByteArrayInputStream(file));
-						embeddedFile.setSubtype(elem.getAttribute("data-content-type") != null ? elem.getAttribute("data-content-type") : "application/octet-stream");
+						String contentType = "".equals(elem.getAttribute("data-content-type")) ? "application/octet-stream" : elem.getAttribute("data-content-type");
+						embeddedFile.setSubtype(contentType);
 						fs.setEmbeddedFile(embeddedFile);
 						String fileName = Paths.get(uri).getFileName().toString();
 						fs.setFile(fileName);
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
@@ -10,7 +10,6 @@ import com.openhtmltopdf.render.BlockBox;
 import com.openhtmltopdf.render.Box;
 import com.openhtmltopdf.render.RenderingContext;
 import com.openhtmltopdf.render.displaylist.PagedBoxCollector;
-import com.openhtmltopdf.util.Util;
 import com.openhtmltopdf.util.XRLog;
 
 import org.apache.pdfbox.pdmodel.PDPage;
--- a/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
+++ b/openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
@@ -10,15 +10,19 @@ import com.openhtmltopdf.render.BlockBox;
 import com.openhtmltopdf.render.Box;
 import com.openhtmltopdf.render.RenderingContext;
 import com.openhtmltopdf.render.displaylist.PagedBoxCollector;
+import com.openhtmltopdf.util.Util;
 import com.openhtmltopdf.util.XRLog;
 
 import org.apache.pdfbox.pdmodel.PDPage;
 import org.apache.pdfbox.pdmodel.common.PDRectangle;
+import org.apache.pdfbox.pdmodel.common.filespecification.PDComplexFileSpecification;
+import org.apache.pdfbox.pdmodel.common.filespecification.PDEmbeddedFile;
 import org.apache.pdfbox.pdmodel.interactive.action.PDAction;
 import org.apache.pdfbox.pdmodel.interactive.action.PDActionGoTo;
 import org.apache.pdfbox.pdmodel.interactive.action.PDActionJavaScript;
 import org.apache.pdfbox.pdmodel.interactive.action.PDActionURI;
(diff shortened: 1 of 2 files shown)


openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
```diff
@@ -224,23 +227,57 @@ private void addUriAsLink(RenderingContext c, Box box, PDPage page, float pageHe
...
+			annotationLink.setAction(uriAct);
+			PDAnnotationContainer annot = new PDAnnotationLinkContainer(annotationLink);
+
+			if ("true".equals(elem.getAttribute("data-embed-file"))) {
+				byte[] file = _sharedContext.getUserAgentCallback().getBinaryResource(uri);
+				if (file != null) {
+					try {
+						PDComplexFileSpecification fs = new PDComplexFileSpecification();
+						PDEmbeddedFile embeddedFile = new PDEmbeddedFile(_od.getWriter(), new ByteArrayInputStream(file));
+						embeddedFile.setSubtype(elem.getAttribute("data-content-type") != null ? elem.getAttribute("data-content-type") : "application/octet-stream");
```
danfickle: I think getattribute returns an empty string if not exists.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
```diff
@@ -224,23 +227,57 @@ private void addUriAsLink(RenderingContext c, Box box, PDPage page, float pageHe
...
+						fs.setEmbeddedFile(embeddedFile);
+						String fileName = Paths.get(uri).getFileName().toString();
```
danfickle, line 259: This won't work for custom protocol handlers reliably. I think it may be better to use another attribute, possibly `download`.

danfickle:
In regard to security, we have the URL resolver which can return null for bad resource attempts. Do we think this is enough? In hindsight the uri resolver should have taken resource type as well as uri, I guess.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
syjer: fixed :) Thank you.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
syjer: You are right, it may fail. Maybe `data-file-name` as an attribute is more telling?

syjer:
about the security implication, I think the current url resolver may not be fine grained enough for this specific case.

I'm not a fan of adding yet another allow list, but as we are embedding resources in the pdf, I think avoiding by default the possibility to let a user embed some arbitrary file from the file system (like the classic `/etc/passwd`) is more sane.

openhtmltopdf-pdfbox/src/main/java/com/openhtmltopdf/pdfboxout/PdfBoxFastLinkManager.java
danfickle: I agree `download` isn't the best attribute name, but it is the standard html5 attribute for this purpose.

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#Attributes

syjer: I didn't know about the download attribute :D . So I could remove "data-embed-file" and use the `download` attribute with a name.

danfickle:
I think we need to have a standard method for all resources (or at least newly added ones), something like:

`builder.useAllowedProtocols(ResourceType.FileEmbed, "http://", "https://");` 

Then we could just add to the enum for each resource, rather than adding methods. The problem is backwards compatibility. Perhaps we could have defaults for each resource type (everything for xml, nothing for file embed, etc) and calling this method would override those defaults.

syjer:
Agree about the using a ResourceType. I wonder if protocol is enough? Maybe adding the possibility  to pass a custom `Predicate<String>` where the parameter is the uri?

For the method name, maybe `defineExternalResourceAccess` is more general? I would guess the more general method would receive 2 parameter: a ResourceType and a `Predicate<String>`. Then we can add an overload for protocols that convert them in predicates internally.

And as you said, we define a sane set of default :).

This way we can centralize the resource access and maybe remove the ad hoc constructor that I've added in https://github.com/danfickle/openhtmltopdf/blob/open-dev-v1/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/BatikSVGDrawer.java#L58

danfickle:
> Agree about the using a ResourceType. I wonder if protocol is enough? Maybe adding the possibility to pass a custom `Predicate<String>` where the parameter is the uri?
> 
> For the method name, maybe `defineExternalResourceAccess` is more general? I would guess the more general method would receive 2 parameter: a ResourceType and a `Predicate<String>`. Then we can add an overload for protocols that convert them in predicates internally.
> 
> And as you said, we define a sane set of default :).
> 
> This way we can centralize the resource access and maybe remove the ad hoc constructor that I've added in https://github.com/danfickle/openhtmltopdf/blob/open-dev-v1/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/BatikSVGDrawer.java#L58

Agree with all of this. The only issue is whether the predicate should be run before or after the uri resolver.

syjer:
> The only issue is whether the predicate should be run before or after the uri resolver.

I would guess after? If you don't mind, I'll open a new PR with only the access control part so we can focus better on that specific feature :)

danfickle:
@syjer, Just letting you know, I'd like to do a release, probably tomorrow. If you are busy, no worries, we can leave this PR until the next release.

Are you otherwise OK with me doing a release?

syjer:
> If you are busy, no worries, we can leave this PR until the next release.

yep, I prefer to let it rest for the next release.

>Are you otherwise OK with me doing a release?

ok on my side :)

syjer:
superseded by #640  :)

PR #512, opened 2020-07-11
remove unused classes/code, raw parameter type, misc. cleanup

syjer:
This PR remove some unused classes, methods, imports, raw parameter type use.

There is still quite a lot of cleanup work to do :).


openhtmltopdf-core/src/main/java/com/openhtmltopdf/render/InlineLayoutBox.java
```diff
@@ -609,7 +609,7 @@ public Dimension positionRelative(CssContext cssCtx) {
     }
     
     // NOTE: Will be List of DisplayListItem when we delete the old renderer.
-    public void addAllChildren(List list, Layer layer) {
+    public void addAllChildren(List<Box> list, Layer layer) {
```
danfickle: I think this should be `? super Box` as it may be a list of `DisplayListItem`. `Box implements DisplayListItem`. I wasn't confident with wildcard generics when I wrote that comment (I still struggle).

danfickle:
I've got to start work on moving the Java2D code to the fast renderer. That way we can delete a massive amount of code.

openhtmltopdf-core/src/main/java/com/openhtmltopdf/render/InlineLayoutBox.java
syjer: fixed :)

danfickle:
Thank you!

PR #518, opened 2020-07-23
Allow to set config for JDKXRLogger in code rather than config file.

danfickle:
Could be useful, maybe?

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/JDKXRLogger.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/JDKXRLogger.java
@@ -21,49 +21,71 @@
  */
 package com.openhtmltopdf.util;
 
-import java.io.IOException;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.stream.Collectors;
 import java.util.logging.Handler;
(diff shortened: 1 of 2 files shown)


syjer:
for sure nicer :).

By the way, I was wondering if it would be beneficial to move the slf4j logger in the core module (and remove openhtmltopdf-slf4j) and use reflection to autoconfigure it if the slf4j api is present on the classpath.

I think it would simplify the use of the library for the end user (and we still keep the possibility to override it in code).
Additionally we could add the support for log4j2 api too.

danfickle:
I'm cautious of using reflection because I'm led to believe it causes all sorts of issues with Java 9 modules (when we upgrade to 11 eventually), osgi modules (#516) and native compilation with Graal VM (which would be nice for serverless). However, if you were to investigate these issues and find them surmountable, I'm not opposed.

P.S. Were you happy with the [logging wiki page](https://github.com/danfickle/openhtmltopdf/wiki/Logging)? Feel free to edit.

syjer:
you are right about the possible issues, I'll have look :).

By the way, I've already looked to do a native compilation with Graal VM some time ago, unfortunately there was an issue with pdfbox, but it's something that maybe has been fixed. If I'm able to make it run, I'll document it .

Cool about the wiki page, it's something that should I've added.

PR #525, opened 2020-08-04
fix #523 remove unused schema/dtd

### LINKED ISSUE #523: useless schemas/dtd in openhtmltopdf-core/src/main/resources/resources  ?
Hi @danfickle , I've noticed in `openhtmltopdf-core/src/main/resources/resources` we have some dtd/schema that are not referenced in the project:

 - `openhtmltopdf-core/src/main/resources/resources/schema/docbook/`
 - `openhtmltopdf-core/src/main/resources/resources/schema/html-4.01/`
 - `openhtmltopdf-core/src/main/resources/resources/schema/xhtml/`

Do they have any use? Removing them does not seems to cause any issue in the build/tests.

syjer:
this PR remove the unused schema/dtd as described in #523 .

This allow a nice decrease of around 200kb (from 1'277'575  to 1'072'995) of the  openhtmltopdf-core jar :)


danfickle:
Awesome, thanks @syjer.

PR #526, opened 2020-08-10
misc cleanup: remove unused method/helper files + add missing generics

syjer:
This pr remove some unused helpers and a whole class (GenerateBigFile.java ).

Additionally, it finally add all the missing generic parameters + some misc. cleanups.


syjer:
Btw, I'm wondering if we could cleanup a little bit the `Configuration` class. There are some features that I wonder if they are even used.

danfickle:
Thanks, I think all the important settings are now set via the builder so we can get rid of `Configuration` class.

syjer:
ok, this will allow to clean quite a lot of code :).

I will provide a separate PR for the refactoring of the Configuration class.

danfickle:
Looks good, merging. Thanks again.

Issue #478, opened 2020-05-12
Dot in li element is not in the middle


mndzielski:
Hi,

I'd like to report that dot in li element is align to top. In browser is align to middle. It's quite annoying.

Another issue is fact that dot is not fully rendering when font-size is large.

HTML example:

```<html>
<head>
<style>
#e1 {
 font-size: 40px;
}

#e2 {
 font-size: 55px;
}
</style>
</head>
<body>

<ul>
<li id="e1">Hello world</li>
<li id="e2">Hello world</li>
</ul>

</body>
</html>
```

And result:

![image](https://user-images.githubusercontent.com/38613340/81703840-62e8ee00-946d-11ea-888d-34e88679f0de.png)

syjer:
hi @mndzielski , I'll have a look

mndzielski:
@danfickle @syjer 
Is there any update on this issue?
![image](https://user-images.githubusercontent.com/38613340/91462740-d2eb7a80-e88a-11ea-82ee-d1e665bf4edc.png)

syjer:
hi @mndzielski currently not, I was side tracked on other stuff, but I'll have look in this period.

mndzielski:
Ok,
so my temporary solution for that issue is:

`
ul {
            list-style: none;
        }
`

` 
        ul li::before {
            content: "\2022";
            font-weight: bold;
            display: inline-block;
            width: 1em;
            margin-left: -1em;
        }`

Hope this may help others

![image](https://user-images.githubusercontent.com/38613340/91463794-198da480-e88c-11ea-8741-25daebfe7e38.png)

syjer:
Currently trying to understand the issue of vertical positioning:

looking at the code in `ListItemPainter.java`:

```java
int y = getReferenceBaseline(c, box) - (int)strutMetrics.getAscent() / 2  - marker.getDiameter() / 2;
```

seems to skew a little bit too much on the top.

Adding some code to visualize it:

```java
c.getOutputDevice().drawDebugOutline(c, box, FSRGBColor.GREEN);
c.getOutputDevice().setColor(FSRGBColor.RED);
c.getOutputDevice().drawRect(x, getReferenceBaseline(c, box), 1000, 1);
c.getOutputDevice().setColor(FSRGBColor.BLUE);
c.getOutputDevice().drawRect(x, getReferenceBaseline(c, box) - (int)strutMetrics.getAscent(), 1000, 1);
```

We can see the issue of the vertical positioning:

[...]

PR #542, opened 2020-08-28
#478 change list decoration placement logic

syjer:
This PR implement what has been done in the issue #478 .

I'll add another test (especially on the RTL side of the equation) before marking the PR as done :).


syjer:
Hi @danfickle , I've marked this PR as ready. I've added a new test which check with various fonts and sizes + the RTL case. Like all the browsers, there will be cases where the marker will not be able to be put inside the default padding, so it will be cut (partially or totally).

To be noted, as written in #478, we need to add the support for `padding-inline-start` + updating the default css for a more out of the box experience for the RTL case, as I added an explicit right padding in the test.

danfickle:
Looks good. Thank you. I'm working on an exciting feature (pr soon) but after that I'll take a look at `padding-inline-start` and friends.

Issue #551, opened 2020-09-11
Thread is stuck during HTML to PDF conversion with nested tables for fairly large HTML


swillis12:
Update 9/23/20: The first two HTML files I attached are misleading and work as expected. I found that the problem is actually occurring only when the height/max-height of the div enclosing the table is set to "auto": 
![image](https://user-images.githubusercontent.com/35236833/94067362-9c7a2000-fda2-11ea-85a1-9c8fae8bfdbb.png)

@syjer Try this new HTML file attached, it should reproduce the issue for you as well. Sorry about the confusion with the previous test cases. FYI running this in my minimal test application using this HTML file was **90,087ms** which is consistent with the results I see inside of my program.
[smaller_test_auto_height.txt](https://github.com/danfickle/openhtmltopdf/files/5271823/smaller_test_auto_height.txt)

Original Issue/info below, the stack trace thread dump info applies to the "smaller_test_auto_height.txt" test case since my proram was using this "auto" height all along (which I wasn't aware of 🤦 ):

Currently our HTML target is email, so we are using a lot of HTML table elements, nested and everything is inline styles. I suspect that I have a thread hanging due to the nested tables. Unfortunately I'm not sure what I can do to work around this issue. I've tried "table-layout: fixed" and assigning column widths as well.

Update: made sure that it is valid XHTML as well using https://validator.w3.org/: [problem_html1.txt](https://github.com/danfickle/openhtmltopdf/files/5211147/problem_html1.txt)

Here is also a smaller test html that is still painfully slow (~1.5-2 minutes). It is the same content, just reduced the number of table rows to 71 for easier visibility: 
[smaller_test.txt](https://github.com/danfickle/openhtmltopdf/files/5211518/smaller_test.txt)

If I enable logging I see infinite messages as follow:
```
com.openhtmltopdf.cascade FINEST:: min-height, relative= 0.0 (0), absolute= 0.0 using base=460.0
com.openhtmltopdf.cascade FINEST:: text-indent, relative= 0.0 (0), absolute= 0.0 using base=0.0
com.openhtmltopdf.cascade FINEST:: min-width, relative= 0.0 (0), absolute= 0.0 using base=0.0
com.openhtmltopdf.cascade FINEST:: min-width, relative= 0.0 (0), absolute= 0.0 using base=0.0
com.openhtmltopdf.cascade FINEST:: height, relative= 23.0 (23px), absolute= 460.0 using base=0.0
[...]

PR #552, opened 2020-09-12
optimize the log formatter

syjer:
Hi @danfickle , this PR try to optimize JDKXRLogger and especially the XRSimpleLogFormatter.

On the JDKXRLogger, we keep a Map so we don't need to go through  `Logger.getLogger` which is kinda slow.

On the XRSimpleLogFormatter side, the biggest win is to avoid calling if necessary:

 - record.getSourceClassName
 - record.getSourceMethodName
 - printing the stacktrace

Both the `record` methods call internally `inferCaller` which it has is own complexity.

To identify which of the parameters are effectively used for the format, I went to the "ugly" route as I'm not aware of any API to check if a parameter is present or not.

I've also aligned the default from the configuration file (see Configuration.valueFor( "xr.simple-log-format") at the bottom).

Generally all this work avoid the quite heavy stacktrace walking, which could be quite long in the cases of deeply nested tables.

Additionally, I've optimized also the formatting of the "diagnostic" message. I've removed the use of String.format. See `LogMessageIdFormat`: this will avoid to call `String.format(MESSAGE_FORMAT_PLACEHOLDER.matcher(logMessageId.getMessageFormat()).replaceAll("%s"), args);` on each message formatting, which is  used quite often if the user use the jdklogger.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/Diagnostic.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/Diagnostic.java
@@ -36,11 +36,7 @@ public class Diagnostic {
     }
 
     public String getFormattedMessage() {
-        if (logMessageId instanceof LogMessageId.LogMessageId0Param) {
-            return logMessageId.getMessageFormat();
-        } else {
-            return String.format(MESSAGE_FORMAT_PLACEHOLDER.matcher(logMessageId.getMessageFormat()).replaceAll("%s"), args);
-        }
+        return logMessageId.formatMessage(args);
     }
 
     public Object[] getArgs() {
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/JDKXRLogger.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/JDKXRLogger.java
@@ -21,14 +21,14 @@
  */
 package com.openhtmltopdf.util;
 
+import java.util.HashMap;
+import java.util.Map;
 import java.util.logging.Level;
 import java.util.logging.Logger;
-import java.util.stream.Collectors;
 import java.util.logging.Handler;
 import java.util.logging.ConsoleHandler;
 import java.util.logging.Formatter;
 import java.util.Arrays;
-import java.util.List;
 
 /**
  * An {@link XRLogger} interface that uses <code>java.util.logging</code>.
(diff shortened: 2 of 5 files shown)


danfickle:
Thanks, there is a lot of impressive work here (or at least it would have taken me a long time to write and debug the string parsing methods).

Merging now.

syjer:
thank you :).

I'll have a look at others possible optimization hotspot, as I'm still wondering the reason of the issue #551 .

testinfected:
We're having what seems to be a race condition in our tests, which run in parallel in different threads :

```
Caused by: java.lang.NullPointerException
	at com.openhtmltopdf.util.JDKXRLogger.getLogger(JDKXRLogger.java:103)
	at com.openhtmltopdf.util.JDKXRLogger.isLogLevelEnabled(JDKXRLogger.java:75)
	at com.openhtmltopdf.util.XRLog.log(XRLog.java:122)
	at com.openhtmltopdf.util.XRLog.log(XRLog.java:85)
	at com.openhtmltopdf.context.StyleReference.getStylesheets(StyleReference.java:253)
	at com.openhtmltopdf.context.StyleReference.setDocumentContext(StyleReference.java:102)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.setDocumentP(PdfBoxRenderer.java:319)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.setDocumentP(PdfBoxRenderer.java:292)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.<init>(PdfBoxRenderer.java:247)
	at com.openhtmltopdf.pdfboxout.PdfRendererBuilder.buildPdfRenderer(PdfRendererBuilder.java:67)
	at com.openhtmltopdf.pdfboxout.PdfRendererBuilder.run(PdfRendererBuilder.java:42)
```

I will put up a test case which reproduces the problem

syjer:
hi @testinfected , 

you are totally right:

code:

```java

public static void main(String[] args) throws Exception {

        int p = 20;
        CountDownLatch latch = new CountDownLatch(p);
        for (int i = 0; i < p;i++) {
            new Thread(() -> {
                latch.countDown();
                try {
                    latch.await();
                    XRLog.log(Level.SEVERE, LogMessageId.LogMessageId0Param.CASCADE_IS_ABSOLUTE_CSS_UNKNOWN_GIVEN);
                } catch (InterruptedException e) {
                }
            }).start();
        }
    }
```

is enough to reproduce the error

syjer:
@testinfected I'll open a new issue so we can track it better.

testinfected:
Thanks @syjer!

syjer:
I've opened the issue #646

PR #554, opened 2020-09-14
log message id format: use static final + add test

syjer:
I noticed that in #552 that I forgot to set as "static" the placeholder. 

Additionally, I've added some tests.


danfickle:
Thanks again. Very concientious!

PR #568, opened 2020-10-12
#549 Fix soft hyphen characters are visible

StephanSchrader:
Fix for issue #549, all invisible characters are filtered and not printed in any way.

This PR replaces: #550


danfickle:
Sensational! This is very nice work. Merging now. Thanks a lot @StephanSchrader.

Issue #592, opened 2020-11-02
SVG images are not rendered with right size


DanielWulfert:
SVG images with either given dimensions or images with no given dimensions are not rendered with the right size. Reason for that is, that _BatikSVGImage.parseLength_ does not handle double values nor handles css units. 
Simple example:
```
<html` xmlns="http://www.w3.org/1999/xhtml">
    <head>
        <meta http-equiv="Content-Type" content="text/xhtml; charset=UTF-8" />
        <title>Test</title>
    </head>
    <body>
        <div id="content" class="content">
            <div id="logo" style="margin-top:3mm;">
                <img src="some_fancy.svg" alt="here comes the SVG" title="" />
            </div>    
            <div id="main" style="font-size:15px;">Some text below the SVG</div>
         </div>
    </body>
</html>
```
The images is rendered correctly, but the space around the image is too big. The text should be exactly below the image and the image should be 3mm away from the top. You can find the following warning messages in the output after the rendering:
`com.openhtmltopdf.general WARNING:: Invalid integer passed as dimension for SVG: 850.39px`
`com.openhtmltopdf.general WARNING:: Invalid integer passed as dimension for SVG: 183.035px`
As mentioned I am using the _BatikSVGDrawer_ from the _openhtmltopdf-svg-support_ plugin. 
Since I have already fixed this issue for my project I will make a pull request later if you like.

danfickle:
Hi @DanielWulfert,

Firstly, thanks heaps for contributing to this project!

I have not used your PR in favor of firing up the CSS parser to deal with unit values as you can see in [commit](https://github.com/danfickle/openhtmltopdf/commit/1e7e5d6f2eab79c3324c1390f1e5cf02959bbe45). I think this will handle more units (font relative units, even percentages) than rolling another parser.

However, without your PR I would not have worked on this issue, so thank you.

PS. You may want to try your SVG with this commit. It gives identical output to Chrome with the simple values I have tried it with but you may find other edge cases with more complex SVGs.

Thanks again,
Dan.

DanielWulfert:
Hi @danfickle 

looks fine to me and thanks for all the efford you put into this project. Great improvement of flying-saucer. Happy to at least contribute a very small part ;)
Keep up the great work.

Regards,
Daniel

kiranduba:
[...]

PR #593, opened 2020-11-02
Issue #592 extended CSS and fractional unit support in BatikSVGImage.parseLength method

DanielWulfert:
These changes will extend the support for fractional numbers and support for absolut CSS units. It will fix issues as mentioned in #592


DanielWulfert:
Obsolete. See #592

Issue #596, opened 2020-11-11
Support for base64 css stylesheet.


Leostat86:
Hello, this is my first time ever to submit an issue. I apologize in advance that i probably did it wrong somehow.

We noticed that using base64 encoded images works great. Which got us thinking, if we base64 encoded our stylesheets we would be able to inspect the html in for instance chrome with CSS and everything. So after we used pebble to do some replacement of texts and such we got out a html with the CSS base64 ecoded.

`<style type="text/css">
@import url("data:text/css;base64,LoolwskegfnhjoiNJEWF..........")
</style>
`

This works perfectly in chrome, but then we tried to use this html in openhtmltopdf we get the following error:
`MalformedURLException: unknown protocol: data`

After inspecting the source code it does not seem that this usage is supported. However this would be a big improvement for us, as this would let us to use chrome as a debugging tool for our html when we are developing it.
I dont know how much effort it would require to add this support, but it would be greatly appreciated.

syjer:
hi @Leostat86 , nice idea. I'll provide a PR :)

PR #597, opened 2020-11-12
#596 generalize data uri support

syjer:
This PR enable the use case described in issue #596 by removing (well, all except the svg data uri handling) the special casing around the use of data uri for images and exposing a protocol handler for the "data" protocol.


danfickle:
Thanks @syjer! Very elegant solution.

PR #598, opened 2020-11-13
misc. util cleanup

syjer:
Hi @danfickle , this PR contains some unused code cleanup.

By the way, I've noticed that we are still using travis.org which will go in read only mode at  the end of the year ( https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom ).

If you want, I can provide a PR to switch to github actions?


danfickle:
Ah, the joy of deleting old code! Thanks. PR to switch would be very welcome. Remember to remove (& replace if possible) build passing image on the readme.

Thanks again.

PR #601, opened 2020-11-16
add github action for CI

syjer:
Hi @danfickle , this PR enable the CI with github actions.

As you can see from https://github.com/syjer/openhtmltopdf/actions?query=workflow%3Abuild it seems to work :)


danfickle:
Thanks mate.

Issue #599, opened 2020-11-13
RuntimeException ocurring on InlineText.setSubstring


LAlves91:
Greetings!

We're using openhtmltopdf in our project. Recently, we've updated from version 1.0.0 to 1.0.4, in order to solve a bug (reported in [issue 420](https://github.com/danfickle/openhtmltopdf/issues/420)).

However, this upgrade caused another bug in our project. Now some documents break with the following stacktrace:

> Caused by: java.lang.RuntimeException: set substring length too long (start = 1, end = 0): InlineText: ()
	at com.openhtmltopdf.render.InlineText.setSubstring(InlineText.java:133)
	at com.openhtmltopdf.layout.InlineBoxing.layoutText(InlineBoxing.java:1114)
	at com.openhtmltopdf.layout.InlineBoxing.startInlineText(InlineBoxing.java:414)
	at com.openhtmltopdf.layout.InlineBoxing.layoutContent(InlineBoxing.java:196)
	at com.openhtmltopdf.render.BlockBox.layoutInlineChildren(BlockBox.java:1220)
	at com.openhtmltopdf.render.BlockBox.layoutChildren(BlockBox.java:1201)
	at com.openhtmltopdf.render.BlockBox.layout(BlockBox.java:1058)
	at com.openhtmltopdf.render.BlockBox.layout(BlockBox.java:973)
	at com.openhtmltopdf.layout.BlockBoxing.layoutBlockChild0(BlockBoxing.java:321)
	at com.openhtmltopdf.layout.BlockBoxing.layoutBlockChild(BlockBoxing.java:299)
	at com.openhtmltopdf.layout.BlockBoxing.layoutContent(BlockBoxing.java:90)
	at com.openhtmltopdf.render.BlockBox.layoutChildren(BlockBox.java:1204)
	at com.openhtmltopdf.render.BlockBox.layout(BlockBox.java:1058)
	at com.openhtmltopdf.newtable.TableRowBox.layoutCell(TableRowBox.java:452)
	at com.openhtmltopdf.newtable.TableRowBox.layoutChildren(TableRowBox.java:206)
	at com.openhtmltopdf.render.BlockBox.layout(BlockBox.java:1058)
	at com.openhtmltopdf.newtable.TableRowBox.layout(TableRowBox.java:95)
	at com.openhtmltopdf.render.BlockBox.layout(BlockBox.java:973)
[...]

PR #604, opened 2020-11-19
#599 Increment _end in trimLeadingSpace

LAlves91:
Increment LineBreakContext ending when trimming leading space. When a tag only has spaces, e.g. an &nbsp; entity, the result was an ending smaller than a start, causing a RuntimeException.


danfickle:
Thanks again @LAlves91.

I think we can close this PR and related issue now. It should be fixed.

LAlves91:
Thank you @danfickle ! Is there a planned date to release the next version?

Issue #608, opened 2020-11-25
Huge amount of +~JF****.tmp files created in the /tmp folder causing system to crash


sureshkumar-ramalingam:
I got the same issue created Huge amount of +~JF****.tmp files created in the /tmp folder causing system to crash

![image](https://user-images.githubusercontent.com/67437117/100239203-1ec7c180-2f57-11eb-8c7a-e85736271aa0.png)

when I look for the solution, I was navigated to the below link.

https://stackoverflow.com/questions/38783010/huge-amount-of-jf-tmp-files-in-var-cache-tomcat7-temp

Any help will be really appreciated.

syjer:
hi @sureshkumar-ramalingam , are you using the java2d backend?

I'm not sure the pdfbox backend is calling Font.createFont.

Could you provide the java code and html example that cause this issue?

Thank you.

sureshkumar-ramalingam:
Hi @syjer It is the pdfbox backend is calling Font.createFont. We are using a custom font file in the HTML.

Thanks,
Sureshkumar

syjer:
Could you provide a little bit more information so we could try to reproduce the error?

For example: the html template, the code that you are using for calling openhtmltopdf.

syjer:
hi @sureshkumar-ramalingam , are you using some svg with text in your html / or simply enabled the svg support?

https://github.com/danfickle/openhtmltopdf/blob/9891184b099a09b103ca1d8633ca5caec5f8ffe2/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/OpenHtmlGvtFont.java#L49 seems to be the only code that is calling Font.createFont in the non java2d path (https://github.com/danfickle/openhtmltopdf/search?q=Font.createFont&type=code ).


If you could show us the java code that you are using for calling this library and a html it will be simpler for tracking this issue (if this library is the main cause).

syjer:
Seems the issue is in the svg support (as reported by  @olayinkasf in comment https://github.com/danfickle/openhtmltopdf/issues/405#issuecomment-733821437).

I think I can provide a fix by handling manually a cache of temporary files with the "hash" as key, so at least for a given font, there is only one instance. On top of adding support for "File" so we can bypass this handling, I hope, in most cases.

PR #611, opened 2020-11-27
remove useless createFont helper

syjer:
This PR is a small cleanup/refactor in the `PdfBoxFontResolver`.

Found this cleanup opportunity while I was exploring the code to understand what could cause the issue in #608 .


danfickle:
Thanks again.

Trying for a release tomorrow.

Issue #405, opened 2019-11-12
Huge amount of +~JF****.tmp files created in the /tmp folder causing system to crash


olayinkasf:
I'm not sure exactly why these files are being created. From s simple google search it seems like they are related to JasperReports and I don't understand why or why. The files seems to also be font files and they contain content related to [Klim Type Foundry](https://klim.co.nz/blog/licensing-information/)

leonorader:
Hi @olayinkasf,
Could you please describe the problem with a bit more details?
What are you trying to achieve? Is there any code samples and input files you could share?

sureshkumar-ramalingam:
I got the same issue created Huge amount of +~JF****.tmp files created in the /tmp folder causing system to crash

![image](https://user-images.githubusercontent.com/67437117/100238883-b842a380-2f56-11eb-8d66-8f0630753f78.png)

when I look for the solution, I was navigated to the below link.

https://stackoverflow.com/questions/38783010/huge-amount-of-jf-tmp-files-in-var-cache-tomcat7-temp

Any help will be really appreciated.

olayinkasf:
I found out that the issue originated from the `BatikSVGDrawer ` that we were using and every time `BatikSVGDrawer#importFontFaceRules` was called these font files were created in the tmp folder. Since we are not using any font in our SVG, I decided to override the method and left it blank.

``` java
    private static final class NoFontSvgDrawer extends BatikSVGDrawer {

        @Override
        public void importFontFaceRules(final List<FontFaceRule> fontFaces, final SharedContext shared) {
        }

    }
```

PR #612, opened 2020-11-29
fix issues #608 / #405 (too many temporary files) by avoiding to call Font.createFont with inputstream

syjer:
Hi @danfickle , I was finally able to understand the issue described in #608, #405 .

This is triggered when using the SVG backend and using custom fonts.

In my case, the sample html is:

```html
<html>
<head>
<style>

@font-face {
  font-family: 'noto';
  src: url(NotoSans-Regular.ttf);
  font-weight: normal;
  font-style: normal;
}

.noto {
  font-family: 'noto', serif;
}

</style>
</head>
<body>
<p>not noto</p>
<p class="noto">noto</p>
</body>
</html>
```

Then by simply iterating:

```java
for (int i = 0; i<100; i++) {
            String input = FileUtils.readFileToString(new File("in.htm"), "UTF-8");

            try (OutputStream os = new FileOutputStream("out.pdf")) {
                PdfRendererBuilder builder = new PdfRendererBuilder();
                builder.useFastMode();
                builder.useSVGDrawer(new BatikSVGDrawer());
                builder.withW3cDocument(W3CDom.convert(Jsoup.parse(input)), "");
                builder.toStream(os);
                builder.run();
            }
            Thread.sleep(1000);
        }
```

You can see the temporary directory slowly fill with a new temporary file during each iteration.

This PR try to ensure that at most only one temporary file is created for a given font (by hashing the byte array), instead of one for each iteration.

Putting as a draft, as I'm not sure it's the best fix.

### DIFF
--- a/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/OpenHtmlGvtFont.java
+++ b/openhtmltopdf-svg-support/src/main/java/com/openhtmltopdf/svgsupport/OpenHtmlGvtFont.java
@@ -4,8 +4,12 @@ import java.awt.Font;
 import java.awt.FontFormatException;
 import java.awt.font.FontRenderContext;
 import java.awt.font.TextAttribute;
-import java.io.ByteArrayInputStream;
+import java.io.File;
 import java.io.IOException;
+import java.math.BigInteger;
+import java.nio.file.*;
+import java.security.MessageDigest;
+import java.security.NoSuchAlgorithmException;
 import java.text.CharacterIterator;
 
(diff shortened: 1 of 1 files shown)


danfickle:
Hi @syjer,

Unfortunately, I'm not brave enough to work with temporary files, so I've created an alternative PR #613 that allows the user to add fonts for SVG as files instead of input streams. More details in that PR.

Thanks anyway for looking at this.

syjer:
hi @danfickle , make sense. I've added a comment in your PR (https://github.com/danfickle/openhtmltopdf/pull/613#issuecomment-735740698) :)

Issue #615, opened 2020-12-03
Table-related infinite loop triggered in versions >= 1.0.2


bsmelo:
Hello!

We found an infinite loop being triggered using openhtmltopdf 1.0.4. Actually, after some tests, we found our code succeeds using openhtmltopdf 1.0.1, but from 1.0.2 onwards it enters the infinite loop. The smallest input we could construct was:
```html
<table style="float:left; margin:1px;">
   <tr></tr>
</table>
<p>NOME</p>
```

Since we have a preprocessing phase using jsoup, the actual input to openhtmltopdf is:
```html
<html>
 <head>
  <style>  body { word-wrap: break-word } table { width: 100% }  </style>
 </head>
 <body>
  <table style="float:left; margin:1px;"> 
   <tbody>
    <tr></tr> 
   </tbody>
  </table> 
  <p>NOME</p>
 </body>
</html>
```

We still couldn't figure out what exactly causes the program to loop, but if we remove any piece of the above input (e.g.: `float:left;` or even `<p>NOME</p>`) it doesn't loop at all. The following code can be used to reproduce the issue:

```java
            ByteArrayOutputStream streamPdfa = new ByteArrayOutputStream();
            PdfRendererBuilder geradorPdf = new PdfRendererBuilder();
            htmlConteudoAjustado = "<THE INPUT ABOVE>";
            
            geradorPdf.withProducer("Nome Produtor");
            geradorPdf.useFastMode();
            geradorPdf.withHtmlContent(htmlConteudoAjustado, "");
            geradorPdf.toStream(streamPdfa);
            geradorPdf.run();
```

We're introducing a local, pre-processing fix, while also debugging and trying to figure out if we can contribute a fix at the openhtmltopdf side.

Thanks for the lib!

syjer:
hi @bsmelo I can confirm the issue.

Looks like a problem inside the Layout algorithm for Inline text.

(More specifically it seems to start from `InlineBoxing.layoutText()` )

syjer:
It seems to remain stuck inside this loop:

https://github.com/danfickle/openhtmltopdf/blob/open-dev-v1/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/InlineBoxing.java#L165

syjer:
the call at https://github.com/danfickle/openhtmltopdf/blob/open-dev-v1/openhtmltopdf-core/src/main/java/com/openhtmltopdf/layout/InlineBoxing.java#L198
[...]

PR #618, opened 2020-12-10
Fix 615 endless loop char breaking

danfickle:
#615 
This will be a two part fix:
+ Make the breaker more robust again by checking expected invariants and logging if not holding.
+ Fix the float clearing algorithm to take account of margins.


danfickle:
I plan to do a release with this on Sunday or Monday.

PR #624, opened 2020-12-20
Update PDFBox and pdfbox-graphics2d dependencies

rototor:
PDFBox just released a new version 2.0.22 and I also had some bugfixes in pdfbox-graphics2d since the last release used here.


danfickle:
Thanks mate. I hope to do a release tomorrow.

PR #627, opened 2020-12-30
Fix font npe

rototor:
Allow fontStyle to be null and don't throw NPE. …

This is a fix for a regression of 1.0.6.


danfickle:
Thanks mate.

PR #628, opened 2020-12-30
Use the PDFBox builtin image direct embedding

rototor:
And don't try to do this ourself. Since some releases PDFBox is able
to not only directly (i.e. without re-encoding) embed images, but
also to directly embed PNG images. TIFF images are supported if
they are in CCITT (i.e. Fax) format.

Note: The PNG direct embedding only works for some PNG encodings,
for details see https://issues.apache.org/jira/browse/PDFBOX-4341

This allows to manually optimise Images using e.g. ImageOptim and benefit
from this optimized images.

If it is not possible to directly embed an image PDFBox falls back to
ImageIO.read() and the LosslessFactory.createFromImage().

In the worst case, this patch may make PDFs bigger, as the
provided image files may be in a worse compression quality
(file size wise) then the LosslessFactory would do when compressing.

But in the best case the PDF writing gets faster, as not every image
needs to be decompressed and compressed again. And the PDFs could get
smaller if the images have been preoptimized.


danfickle:
Fully agree. Thanks again.

PR #638, opened 2021-01-19
Implement external access control based on previous PR

danfickle:
#528 #509 @syjer @vader

Based on #528 so we can implement #509.

Comments welcome.

Needs a couple more tests.


syjer:
hi @danfickle , thank you for taking over the two PR. I didn't had enough time/focus for completing them...

Looking at the code I think all the use cases are covered.

Issue #636, opened 2021-01-15
Support for PDF/A-3


vader:
A request to add support for PDF/A-3 which only differs from PDF/A-2 in one regard: it allows embedding of arbitrary file formats (such as XML, CSV, CAD, word-processing documents, spreadsheet documents, and others) into PDF/A conforming documents. 
I noticed there is some A-3 support in PDFAFlavour.PDFA_3_U. 

I am looking to embed an XML file in the PDF as an attachment. 

I would be happy to help with some pointers and assistance.

danfickle:
Hi @vader,

We have PR #508 (also #528) which I think may do what you are looking for. It is not yet ready for merge, but I'll have a look at what it will take to get it merged.

danfickle:
Support now merged into main branch.

[File embed documentation](https://github.com/danfickle/openhtmltopdf/pull/640#issuecomment-770819471).

PR #640, opened 2021-01-21
Implementation of file-embed link.

danfickle:
#508 #509 #636

Based heavily on code by @syjer whom I'm indebted to. Thanks.

Still todo:
+ [x] Prevent duplicate file embeds. Possibly create a map of uris to PDComplexFileSpecification and reuse if encountered again. I have to peruse the PDF spec to see if this is allowed.
+ [x] Logging on fail.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/LogMessageId.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/util/LogMessageId.java
@@ -111,6 +111,7 @@ public interface LogMessageId {
         LOAD_COULD_NOT_INSTANTIATE_CUSTOM_XML_READER(XRLog.LOAD, "Could not instantiate custom XMLReader class for XML parsing: {}. " +
                 "Please check classpath. Use value 'default' in FS configuration if necessary. Will now try JDK default."),
         LOAD_UNABLE_TO_LOAD_CSS_FROM_URI(XRLog.LOAD, "Unable to load CSS from {}"),
+        LOAD_COULD_NOT_LOAD_EMBEDDED_FILE(XRLog.LOAD, "Was not able to load an embedded file for embedding with uri {}"),
         LOAD_PARSE_STYLESHEETS_TIME(XRLog.LOAD, "TIME: parse stylesheets {}ms"),
         LOAD_REQUESTING_STYLESHEET_AT_URI(XRLog.LOAD, "Requesting stylesheet: {}"),
         LOAD_UNRECOGNIZED_IMAGE_FORMAT_FOR_URI(XRLog.LOAD, "Unrecognized image format for: {}"),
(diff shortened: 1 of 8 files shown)


syjer:
Thank you @danfickle , I'll close the old PR :)

danfickle:
OK, here is a valid file embed link for PDF/A3 standards:
````html
   <a href="file-embed.html"
      download="source.html"
      data-content-type="text/html"
      title="File embedded"
      relationship="Source">

      Link to source code

   </a>
````
The attributes are:
+ **href**: The file source. Required.
+ **download**: The file name as presented in the PDF viewer. Required.
+ **data-content-type**: The mime type of the embedded file. Defaults to `application/octet-stream`.
+ **title**: Optional short description of the file. Strongly recommended for PDF/A3 documents.
+ **relationship**: Should only be used in PDF/A3 documents. The relationship of the embedded file to the link. **Must** be one of the following:
  + `Source` - Example: source markup of the document.
  + `Supplement`
  + `Data` - Example tabular data.
  + `Alternative` - Example audio providing an alternative view of the material.
  + `Unspecified` - Fallback catch all.

**NOTE:** If the `relationship` attribute is present the files modification date is set to the servers current date (`Calendar.getInstance()`). The modification date is required by PDF/A3.


Useful links in writing this implementation:
+ [PDF/A-3u as an archival format for Accessible mathematics](https://arxiv.org/pdf/1406.6126.pdf) (large PDF)
+ [Library of Congress on PDF/A3](https://www.loc.gov/preservation/digital/formats/fdd/fdd000360.shtml)

**Comment by houssainy** [comment]
@danfickle 
I want to embed an XML file to the generated PDF/A3 file but I don't really want the <a> tag to appear in the generated PDF. So is there a way to hide the `<a>` tag?

I tried multiple ways to hide the <a> tag but all of them were failing. Also i tried to have an empty <a> tag 
```
 <a href="file-embed.html"
      download="source.html"
      data-content-type="text/html"
      title="File embedded"
      relationship="Source">
   </a>
   ```
   but this also didn't work and generated PDF didn't contain the expected attachment.

maciejtoporowicz:
Hello, does it work in fast mode? It only works for me in slow mode.

In fast mode I can see these logs:
```
com.openhtmltopdf.load WARNING:: URI file:///<path> with type FILE_EMBED was rejected by resource access controller
com.openhtmltopdf.load WARNING:: Was not able to load an embedded file for embedding with uri file:///<path>
```

The PDF is created and the link is rendered, but it's not clickable.

maciejtoporowicz:
Ok, figured it out. The `DefaultAccessController` used in `com.openhtmltopdf.swing.NaiveUserAgent` rejects embedding files.
I don't really understand this mechanism, but I found out that it's configurable. I created my own "AccessController":
```
private static class AllowAllAccessController implements BiPredicate<String, ExternalResourceType> {
  public boolean test(String uri, ExternalResourceType resourceType) {
    return true;
  }
}
```
and then configured the pdf renderer to use it:
```
new PdfRendererBuilder()
    .useExternalResourceAccessControl(new AllowAllAccessController(), ExternalResourceControlPriority.RUN_BEFORE_RESOLVING_URI)
    .useExternalResourceAccessControl(new AllowAllAccessController(), ExternalResourceControlPriority.RUN_AFTER_RESOLVING_URI)
    (...)
    .run();
```

Now it works fine.

PR #658, opened 2021-02-21
Add feature: target-text() for content property capabilities

BenjaminVega:
Implementation of target-text() for content property. (https://www.w3.org/TR/css-gcpm-3/#target-text)


danfickle:
Awesome! Works well and with test. Can't argue with that! Thanks a lot @BenjaminVega.

BenjaminVega:
Thanks you for this great library! 👏

PR #678, opened 2021-03-26
Upgrade  Batik Version  to 1.14 (CVE-2020-11987)

rototor:
Details see here https://xmlgraphics.apache.org/security.html and https://nvd.nist.gov/vuln/detail/CVE-2020-11987

I don't think that this is super urgent.  But should be upgraded anyway


danfickle:
I'm back. Sorry to everyone for the extended outage! Thanks a lot @rototor.

PR #685, opened 2021-04-03
Do not use org.apache.pdfbox.util.Charsets, use StandardCharsets

rototor:
In PDFBox 3.0 org.apache.pdfbox.util.Charsets has been removed,
as it has been obsoleted by StandardCharsets which exists since
Java 1.7


danfickle:
Thanks again! Will research pdfbox 3 and reply in issue at some stage (got a lot of issues to go through).

PR #701, opened 2021-05-14
Fix log warnings, cleanup tests

danfickle:
+ Fix for serious log warnings from PDFBOX saying we are doing the wrong things.
+ General clean up of some tests.


danfickle:
I actually started this PR because I was seeing warnings like this in my logs:

> WARNING: The end of the stream doesn't point to the correct offset, using workaround to read the stream, stream start position: 413, length: 2083, expected end position: 2496

It turns out this was caused by some proof PDFs having line-endings turned from lf to crlf when downloaded from Github onto my Windows machine. I've since added a .gitattributes file that marks .pdf as binary so this won't happen again but didn't fix the problematic files in my working copy.

SO came to the rescue with [git commands in this answer to fix line-endings](https://stackoverflow.com/a/42135910/6522707).

PR #711, opened 2021-05-29
Issue #364 Implement footnotes at bottom of page

danfickle:
**Now documented on [the wiki under footnotes](https://github.com/danfickle/openhtmltopdf/wiki/Footnotes).**

Fixes #364.

Finally ready!

<s>Need tests and fixes for:</s>

Performance:
+ [x] Many footnotes.
+ [x] Footnotes in amongst many lines of text.
+ [x] Footnotes nested at great depth.

Blockers:
+ [x] Don't break Java2D single page output. - Footnotes inside single page output will place them between virtual pages (ie. nonsensical but should not crash).
+ [x] Java2D multi-page output. - Works correctly.
+ [x] Make sure PDF/UA content is not broken. - Footnotes are NOT supported in PDF/UA.

More:
+ [x] Counter reset on page-break - Not supported for this PR.
+ [x] Links in documents with footnotes are not targeting correctly.
+ [x] Footnotes called from positioned content (absolute, relative, fixed) - Absolute and relative works but physical ordering of footnotes may be different to counter order. Banned from fixed area.
+ [x] Footnotes called from page margins - Banned (treated as normal content) and warning logged.

Inside footnotes:
+ [x] Inline content
+ [x] Block content
+ [x] Inline-block content
+ [x] Table content - paginated - unsupported (incorrect output) but does not crash, OOM, etc.
+ [x] Table content not paginated
+ [x] Replaced content (images)
+ [x] Floated content - is contained by each footnote body.
+ [x] Absolute positioned content - will be contained by footnote area (positioned relative to).
+ [x] Fixed positioned content - will escape footnote area and be positioned as normal fixed boxes.
+ [x] Transformed content
+ [x] Footnotes inside footnotes - Warning emitted and treated as normal content.

As footnote element:
+ [x] Inline content - will be displayed as block box regardless.
+ [x] Block content
+ [x] Inline-block content - Not supported, treated as normal content.
+ [x] Table content - paginated - Not supported, treated as normal content.
+ [x] Table content not paginated - Not supported, treated as normal content.
+ [x] Replaced content (images) - Images may not be used as footnote element but can be wrapped in another element such as div. Images can be used as the pseudo footnote-call or footnote-marker via the content property.
+ [x] Floated content - Footnotes can only be float: bottom.
+ [x] Absolute positioned content - will be contained by footnote area (positioned relative to) with footnote mark.
+ [x] Fixed positioned content - will escape footnote area and be positioned as normal fixed boxes, but without footnote mark.
[...]

PR #722, opened 2021-06-15
Upgrade PDFBox  and PDFBoxGraphics2D

rototor:
PDFBox just got a new release and PDFBoxGraphics2D allows now for parts of a drawing to be marked using BMC/EMC and BDC/EMC


danfickle:
Thanks @rototor. I was hoping to finish footnote support (#711) but since this is a security release of PDFBOX, I'll try to do a release with what we've got first (probably today or tomorrow).

PR #727, opened 2021-06-23
Fast renderer should support initialPageNo

fanthos:


danfickle:
Thanks a lot @fanthos! I've added a test and builder support and will merge when CI is finished. Thanks for contributing.

Issue #635, opened 2021-01-13
Issue with table border width and pagination


axd3:
Hi,

first of all, thank you for this awesome library.

I've encountered some problems with tables I just can't seem to fix myself.

1. When I create a table where borders cross there seems to be an issue with the border width. It appears to be thicker for one cell after the junction (see screenshot 1). I tried it with different pdf-readers and browsers, and although it appears to get better the more you zoom out, it really stands out when zoomed in. Maybe I'm doing something wrong with the css? Or it is a bug? The lines don't appear to have different width when I use either vertical or horizontal lines, only when they cross, and just for one cell in each direction.
2. Pagination for larger tables works fine, my headers repeat on each page and everything looks good. However, if I use `page-break-after: avoid;` in the css, and a page break is avoided by this rule, the headers disappear. The borders for them are still there but the text is missing (see screenshot 2). When the `page-break-after: avoid;` rule does NOT apply for one page, everything looks good again.

I created a repository [here](https://github.com/axd3/pdf-table-test) where you can reproduce the beaviour. Just run the PdfTableTest.java and look at foo.pdf. Any help would be appreciated!

Screenshot 1
![Screenshot_20210113_135350_edit](https://user-images.githubusercontent.com/53512923/104464495-47615480-55b3-11eb-97d3-c1549557efc4.png)

Screenshot 2
![Screenshot_20210113_151332](https://user-images.githubusercontent.com/53512923/104464027-c30ed180-55b2-11eb-8f0c-a78aa2ce9b61.png)

Thank you!

syjer:
the thickness issue may be related to also this issue https://github.com/danfickle/openhtmltopdf/issues/310 and maybe even https://github.com/danfickle/openhtmltopdf/issues/291 ?

PR #755, opened 2021-09-07
Fixes #752 - Use non-bevelled borders for collapsed table cells.

### LINKED ISSUE #310: Table border line thickness is inconsistent
I have a problem when rendering html to PDF using openhtmltopdf.  As below picture, The border color of the first column and the last column of the table is lighter than other columns. Has anyone encountered this problem?
image

### LINKED ISSUE #752: table borders: single border display bug with border-collapse: collapse
When some cell borders are hidden (as 2, 3 and 4 cells in example)
There's buggy behavior of table borders in generated pdf. 
It's usually ok with default 100% zoom level. 
But if it's zoomed in or zoomed out, some borders looks very unstable (cell 3 in example)
Depending on zoom level it can be thinner or thicker than other borders.

example is created with openhtmltopdf sandbox. 
html 

image

image
image

markup
```
<table>
  <tr>
    <td>1</td>
    <td class="no-border-right">2</td>
    <td class="no-border-left no-border-right">3</td>
    <td class="no-border-left">4</td>
  </tr>
  <tr>
    <td>5</td>
    <td>6</td>
    <td>7</td>
    <td>8</td>
  </tr>
</table>
```

css 
```
table {
  border-collapse: collapse;
  border: 1px solid black;
}
table td {
  border: 1px solid black;
  width: 100px;
}
table td.no-border-right {
  border-right: none;
}
table td.no-border-left {
  border-left: none;
}
```

P.S. even without border collapse, with cellpadding="0" and some tricks
 like display only right and bottom cell borders and for all table left and top borders
 (to make sure that there's indeed only single border everywhere)
 border thickness is displayed buggy when zoom in/zoom out

danfickle:
This stops them being anti-aliased which can look awful. I was able to piggy back off someone's work to remove bevelling on collapsed table cell borders from 14 years ago. It must have broke a long the way before automated testing was introduced.

Fixes #752, fixes #310.

Unfortunately this will break tests incorporating collapsed tables with table cell borders because of the tiny difference at the removed bevel. Any objections to merging?

P.S Release this week is planned.


syjer:
nice, I would guess it also fix https://github.com/danfickle/openhtmltopdf/issues/635 ?

danfickle:
Replaced with #761 which adds a property to explicitly disable bevelling, and thus avoids broken tests.

danfickle:
I'll remove the backing branch for this PR as #761 is essentially an extension of the changes.

PR #768, opened 2021-09-25
update github action from java 17-ea to 17

syjer:
looks like that the 17-ea download the stable version anyway, but I guess at least it's more clear java17 is out :)


danfickle:
Thanks @syjer.

PR #773, opened 2021-10-02
#771 Initial work on supporting multi-line text in target-text

### LINKED ISSUE #771: Table of Content does not support multi-line title
Hi!

Thanks for this amazing library. I am trying to generate a table of content using ´target-text´ text that I added months ago. like this:
`#five::after { content:  target-text(attr(href)) leader(dotted)  target-counter(attr(href), page); }`
`<a id="five" href="#title" style="display: inline-block;width: 100%;"></a>`
`<h1 id="title" style="font-size: medium">Title 1</h1>`
Resulting in :

Title 1 . . . . . . . . . . . . . . . . . 2

 It just happen that the targeted text was a very long and then I started having errors.

My expectations were to see the title line being broken into two or more lines but it did not happen. Actually it throw an exception saying something about a negative length. (java.lang.NegativeArraySizeException: -78). After investigating the exception, this is thrown while calculating how many dots to append before the page number . That happens because the Text line is longer than the space of the page. Then I realised I needed to break the lines. I tried it  like this:
image

As you can see, the title itself is not being broken into lines. I have tried adding '\A' and white-space: pre in the `target-text` function to break the line manually  when it extract the tex but it just print it literarily. 

Do you have any suggestion ?

Thanks

danfickle:
Fixes #771 by laying out `target-text` content at layout instead of waiting until render time. This means we can handle multi-line text.

Working well. This PR also:
+ Allows attributes other than `href` to be used with `target-text`.
+ Bans the use of two `leader` functions on the one line to avoid stack overflow as they fight.
+ Tests complex `target-text` usage including surrounding or surrounded by `leader` functions.


BenjaminVega:
👏

Issue #792, opened 2021-12-14
target-counter decimal only


yudieg:
When using CSS to show page number with lower-roman it shows as decimal.

This rendered as decimal.
` content: leader(dotted) target-counter(attr(href), page, lower-roman);`

syjer:
hi @yudieg , I've created a pull request to add the requested feature.

yudieg:
@syjer  thank you!

PR #793, opened 2021-12-17
#792 target counter list style

syjer:
Hi @danfickle , this PR handle the third parameter (counter style) in the target-counter function as requested in issue #792 .


danfickle:
Thanks @syjer - I'm doing a mini security review (inspired by current events) at the moment so trying not to get involved in other issues. I'll try to do a release with this and extra security tests in the next couple of weeks.

PR #794, opened 2021-12-19
Upgrade to current PDFBox and PDFBox-Graphics2D version.

rototor:


rototor:
Oh no! PDFBox-Graphics2D did not yet sync to maven central in time :( In some hours it should finally be fetchable from maven central.

danfickle:
Thanks again @rototor. Diff image for test attached for posterity.

472 test diff image

PR #805, opened 2022-02-04
Improve error reporting for missing font

qligier:
This modifies the PdfBox renderer to throw a new exception (`FontNotFoundException`) in the case of a missing font instead of trying to access an empty font list and throw `IndexOutOfBoundsException`. A check is added in `PdfBoxFastOutputDevice.setFont(FSFont)` to assert the font description isn't null. The methods `AbstractOutputDevice.drawText(RenderingContext, InlineText)` and `ListItemPainter.drawText(RenderingContext, BlockBox, IdentValue)` are modified to set the font specification before the font itself to make it available in the `PdfBoxFastOutputDevice.setFont(FSFont)` call.

The exception thrown goes from:
`java.lang.IndexOutOfBoundsException: Index 0 out of bounds for length 0`
to 
`com.openhtmltopdf.pdfboxout.fontstore.FontNotFoundException: No font for the following specification has been found: Font specification:  families: [arimo, serif] size: 320.0 weight: normal style: normal variant: normal`

I've tested it in some of my documents and all tests are passing.

### DIFF
--- a/openhtmltopdf-core/src/main/java/com/openhtmltopdf/render/AbstractOutputDevice.java
+++ b/openhtmltopdf-core/src/main/java/com/openhtmltopdf/render/AbstractOutputDevice.java
@@ -71,8 +71,8 @@ public abstract class AbstractOutputDevice implements OutputDevice {
        
         if (text != null && text.length() > 0) {
             setColor(iB.getStyle().getColor());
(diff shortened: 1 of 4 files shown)


danfickle:
Looks good! Thanks a lot @qligier

PR #840, opened 2022-05-18
fix when enable accessibility , can only create two pages, the rest of

pdfnewbie:
pages showing blank page.
Temp fix only for RDC project.


pdfnewbie:
The fix is not final.

PR #841, opened 2022-05-26
Adding JPEG autorotation

smahnken:
This change adds automatic image rotation for JPEG images with orientation EXIF data.

Almost all cell phone cameras store the image data in an un-rotated state, and write EXIF data to indicate "which side should be up".  All modern browsers understand these EXIF tags and automatically orient the image correctly.  Thus, when the pictures are viewed in a web page they appear correct.  When the page is output to PDF, however, the image is rotated.

This change modifies the pdfboximage class to read the EXIF data (if it exists) and rotate the image data if necessary.  This results in a "rotate on load" behavior (i.e. before the resource cache), so the rotation only needs to be done once, even if the image appears multiple times in the document.

This change utilizes the apache-commons imaging library (which is currently in alpha-3) to read the EXIF data.  

Please go easy on me.  This is my first pull request.  Let me know if I've done something incorrectly, and I'll do my best to comply.


Dunemaster:
hi, @danfickle 

Any plans to merge the pr?

Thanks

justvdv:
Hi @danfickle @Dunemaster @smahnken 

Any update on this PR? Running into the same issue at the moment.

Zorobay:
We are also experiencing this problem, and would love this solution to be merged in! What can we do to make that happen?

PR #870, opened 2022-09-21
Fixed typo in method name 'accessbility'

furnivall:
Just a quick and easy one - love the service!


zacharysiegel:
See https://github.com/openhtmltopdf/openhtmltopdf/pull/56

PR #911, opened 2023-02-15
Add Signature Field

burka:
Fixes #681

This adds the signature field as requested in #681 :-) 

An example form-signature-field.html is included.

I couldn't get the formatting perfect, hope it's Ok. 

I had to pass down the RenderingContext due to NPEs while setting up the fonts.

PR #913, opened 2023-02-28
Forms on multi page pdfs

### LINKED ISSUE #914: Inputs / Form-Elements on Multi-Page-PDFs on the not-last-page are broken
See PR #913 for example of a failing form. The coordinates are always calculated relative to the last page of the PDF which wont work.

burka:
Fixes #914
Forms on PDFs with multiple pages that are not on the last page are broken. They get the last pages RenderingContext and that will calculate wrong coordinates. With this PR it will use the right page, fixing the RenderingContext to be correct and not attach a changing object to the PdfBoxForm.Control would be even better, but more intrusive.


burka:
PR #911 is contained in this PR and should be merged before :-)

burka:
Fixes #914

zacharysiegel:
FYI, for anyone visiting this page in the future, another version of this PR was merged into the new openhtmltopdf repository [here](https://github.com/openhtmltopdf/openhtmltopdf/pull/37)

Issue #917, opened 2023-03-08
Potential bug - NPE in COSArray.add with big HTML document in PDF A_1_A mode and Arial font


CedricMtta:
Hello,

First of all, thanks for the work performed on this lib, it's really appreciated.

I work on a project using openhtmltopdf to generate PDF A_1_A compliant from quite big HTML files.
We ran into the following stacktrace in some cases:
`java.lang.NullPointerException: Cannot invoke "org.apache.pdfbox.pdmodel.common.COSObjectable.getCOSObject()" because "object" is null

	at org.apache.pdfbox.cos.COSArray.add(COSArray.java:62)
	at com.openhtmltopdf.pdfboxout.PdfBoxAccessibilityHelper.finishNumberTree(PdfBoxAccessibilityHelper.java:858)
	at com.openhtmltopdf.pdfboxout.PdfBoxFastOutputDevice.finish(PdfBoxFastOutputDevice.java:908)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.writePDFFast(PdfBoxRenderer.java:674)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.createPdfFast(PdfBoxRenderer.java:564)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.createPDF(PdfBoxRenderer.java:490)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.createPDF(PdfBoxRenderer.java:427)
	at com.openhtmltopdf.pdfboxout.PdfBoxRenderer.createPDF(PdfBoxRenderer.java:409)
	at com.openhtmltopdf.pdfboxout.PdfRendererBuilder.run(PdfRendererBuilder.java:46)
	at com.openhtmltopdf.pdfa.testing.PdfATester.run(PdfATester.java:95)
	at com.openhtmltopdf.pdfa.testing.PdfATester.cannotRenderInArial(PdfATester.java:153)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
[...]

PR #920, opened 2023-03-09
:bug: Link newly created accessibility object to existing one while ensuring ancestor tree

CedricMtta:


CedricMtta:
Please see https://github.com/danfickle/openhtmltopdf/issues/917

PR #941, opened 2023-09-01
pdfbox 3

hughsimpson:
With the [3.0.0 release](https://github.com/apache/pdfbox/releases/tag/3.0.0) of pdf box, current openhtmltopdf releases are incompatible. Adding some final compatibility tweaks.


openhtmltopdf-objects/src/main/java/com/openhtmltopdf/objects/pdf/PdfDrawerBase.java
```diff
@@ -33,12 +34,10 @@ protected PDFormXObject importPageAsXForm(RenderingContext ctx, Element e,
         PDFormXObject pdFormXObject = map.get(url);
         if (pdFormXObject == null)
         {
-            try (InputStream inputStream = new URL(url).openStream())
-            {
-                PDDocument document = Loader.loadPDF(inputStream);
-                pdFormXObject = layerUtility
-                        .importPageAsForm(document, pdfpage - 1);
-            }
+            File file = new File(new URL(url).toURI());
```
hughsimpson, line 37: Not sure if there are things this might break.. the input stream api has gone, and the alternative is to read the full byte array, which seems too heavy... OTOH maybe `new URL(url).openStream()` does some lookup resolution that's invalid for `new File(new URL(url).toURI())`? I don't really know this part of the java API too well, so any 👀 would be appreciated

enstulen:
Awesome! @danfickle Since 3 is officially released, will this be updated as well?

zacharysiegel:
FYI, the new [openhtmltopdf repository](https://github.com/openhtmltopdf/openhtmltopdf) incorporates PdfBox 3.

hughsimpson:
I very rarely have a pr merged after more than a year, and the fork has been published and updated frequently, so I now consider this dead

